PR #3, opened 2014-10-27
Completing exercises for Chapter 1.

NormanLittle:


src/com/selftest/conversion/InchesToMetresTable.java
```diff
@@ -0,0 +1,18 @@
+package com.selftest.conversion;
+import java.util.stream.IntStream;
+
+public class InchesToMetresTable {
+	private static final double INCHES_TO_METRES_RATIO = 39.37;
+	
+	public static void main( String[] args ) {
+		
+		// 12 feet => 12 x 12 inches => 144 inches.
+		IntStream.rangeClosed( 1, 144 ).forEach(
```
SandyThomson, line 10: Neat.

NormanLittle, line 10: @SandyThomson  - I was actually hoping I could have used a nested IntConsumer (using the andThen() method) to do the "12th row" check but I couldn't find a way to get it working.  Have you come across this before?  Any ideas?

PR #14, opened 2014-11-04
Chapter 2 Try This and Self Test

codavex:


src/ch02/q10_simple.java
```diff
@@ -0,0 +1,39 @@
...
+      // horribly inefficient, testing against numbers that aren't prime
+      for(div = 2; div < num; div++) {
+        if(num % div == 0 ) {
+          // proven otherwise!
+          isPrime = false;
+        }
+      }
+
+      // Report ..
+      if(isPrime == true) {
```
utopia-2, line 30: This is unnecessary, the result of a relational operator is itself just a boolean, so you can `if( isPrime ) {`.

PR #16, opened 2014-11-05
Red - Chapter 2

utopia-2:
Chapter 2 exercises for Java Training.

### DIFF
--- a/answers.md
+++ b/answers.md
@@ -1,0 +1,44 @@
+## Chapter 2 Answers
+
+   1. Java strictly specifies the range and behaviour of primitive types to
+      preserve portability across different hardware.
+   
+   2. Java's character type is unicode (UTF8 for single values, interestingly UTF16 for
+      arrays and Strings to guarantee fixed-width contiguous memory usage).
+      This differs from some other languages that typically use fixed-width 8
+      bit characters.
+      
+   3. False, a boolean value can only be assigned true/false literals, or the result of
+      a logical or relational operator.
+      
+   4.
+   
+```java
+    System.out.println( "One\nTwo\nThree" );   
+```
+
+   5. `i` is not declared, `sum` is not initialised, and an attempt is made to use
+      `sum` outside of the scope is was declared in.
+      
+   6. The prefix form of increment operator increments the value and returns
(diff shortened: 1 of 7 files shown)


answers.md
```diff
@@ -0,0 +1,44 @@
+## Chapter 2 Answers
+
+   1) Java strictly specifies the range and behaviour of primitive types to
+      preserve portability across different hardware.
+   
+   2) Java's character type is unicode (UTF8 for single values, interestingly UTF16 for
+      arrays and Strings to guarantee fixed-width contiguous memory usage).
+      This differs from some other languages that typically use fixed-width 8
+      bit characters.
```
utopia-2, line 9: I'm wrong, Java chars are _all_ in UTF-16 - representing values outside of this range is done by allowing `char`s to include a _supplementary_ `char` (thus a single `char` var can be composed of two UTF-16 characters for values `0x10000` to `0x10FFFF`). The point about iteration over fixed-width memory is still essentially sound, though the Java definition simply didn't use UTF-8 for backwards compatibility reasons.

answers.md
```diff
@@ -0,0 +1,44 @@
...
+```
+
+   8) `byte` and `short` are promoted to `int` in an expression.
+   
+   9) A cast is needed when conversion between types may cause a loss of
+      information (or when people don't understand `<T extends SomeClass>`
+      parameterization, use the `SomeClass` interface and cast it to the
+      desired type later on `</rant>`).
+      
+  10) See `PrimeNumbers.java`.
```
utopia-2, line 40: Came up with this sieved variation when trying to do it faster (it's actually a lot slower for large primes, because of the allocation cost, perhaps a mark and sweep method would be better?):

``` java
final int limit = 1_000_000;

List<Integer> primes = new LinkedList<Integer>();
LinkedHashSet<Integer> range = new LinkedHashSet<Integer>();
range.add(2);
for( int i = 3; i < limit && range.add(i); i += 2 );

for( int p, i; range.size() > 0; range.remove(p), primes.add(p) )
  for( p = range.iterator().next(), i = p * p; i <= limit; range.remove(i), i += p );

primes.forEach(System.out::println);
```

PR #17, opened 2014-11-05
Calculating prime (Optimus?)

dickyince:


PrimeCalculator.java
```diff
@@ -0,0 +1,42 @@
...
+	  if ( prime ) {
+	    primes.add( i );
+	  }
+	}
+	return primes;
+  }
+
+  public static void main( String[] args ) {
+    int start = ( args[0] != null ? Integer.parseInt( args[0] ) : 2 );
+	int end = ( args[1] != null ? Integer.parseInt( args[1] ) : 100 );
```
utopia-2: Java arrays will `ArrayIndexOutOfBoundsException` on you if you try and access values outside of the array bounds, not return `null`. A safer alternative would be:

``` java
int start = args.length > 0 ? Integer.parseInt(args[0]) : 2;
```

answers.md
```diff
@@ -0,0 +1,25 @@
+## rpi answers
+
+1. so all operations are type-checked by the compiler, preventing errors in code
```
utopia-2, line 3: That doesn't really answer why the range and behavior of primitives is strict. Compile-time type checking doesn't do range checking on operations. I think the question means, for example, _"why does java force ints to be signed 32 bit  across all implementations instead of defaulting to the most compatible type available in the underlying hardware"_.

PrimeCalculator.java
```diff
@@ -0,0 +1,42 @@
...
+  public PrimeCalculator( int start, int end ) {
+    this.start = start;
+	this.end = end;
+  }
+
+  public List<Integer> calculatePrimes() {
+    ArrayList<Integer> primes = new ArrayList<Integer>();
+	for( int i = start; i <= end; i++ ) {
+	  if ( i < 2 ) {
+	    continue;
```
utopia-2, line 16: Instead of checking for `i < 2` on _each_ iteration, why not initialize like:

``` java
for ( int i = start > 2 ? start : 2; i <= end; i++ ) {
```

Or specify the constraint in the constructor.

PrimeCalculator.java
```diff
@@ -0,0 +1,42 @@
...
+      }
+      boolean prime = true;
+	  for ( int j = 2; j <= ( i / 2 ); j++ ) {
+	    if ( i % j == 0 ) {
+		  prime = false;
+		  break;
```
utopia-2, line 22: Instead of needing to keep track of whether it's a prime with `prime`, it might be simpler to label the outer loop and use a `continue LABEL` statement. Then it can simply be:

``` java
PRIMELOOP:
for ( int i = start; i <= end; i++ ) {
  //...
    if ( i % j == 0 )
      continue PRIMELOOP;
  }
  primes.add( i );
}
```

Unless you feel like labeled `continue`s are too close to GOTO's for comfort.

PrimeCalculator.java
```diff
@@ -0,0 +1,42 @@
+import java.util.*;
+
+class PrimeCalculator {
+  private int start;
+  private int end;
+
+  public PrimeCalculator( int start, int end ) {
+    this.start = start;
+	this.end = end;
```
utopia-2, line 9: Might be worth adding in some basic constraints here, like `start` should be 2 or greater, and `end` should be greater than or equal to start. As it is, the user doesn't get a warning if they mess up the start/end ordering, just an empty list.

PrimeCalculator.java
dickyince, line 16: That brakes the code as the loop would never start, what if I wanted to know the primes between 1 and 50?

utopia-2, line 16: The loop would do exactly as it does with your `if ( i < 2 ) continue;` except it wouldn't have to waste cycles on that condition (and all following iterations checking that condition).

PR #19, opened 2014-11-05
mcm ch2 exercises

mocathe1st:
Chapter 2 Exercises


mcm/src/mcm_ch2/ch2answers.markdown
```diff
@@ -0,0 +1,69 @@
+
+# Java A beginner's guide - chapter 2 Answers
+*(Markdown doc - best viewed on GitHub)*
+
+**Q1** Java variables are strongly types to promote compatibility, help prevent errors and increase reliability.
+
+**Q2** Java's charcter type is unicode, and is an unsigned 16-bit data type, allowing for a ranfe from 0 to 65,536. This means that Java's character type is capable of storing the majority (if not all) of the world's character sets. This may be inefficient for commonly used character sets, but is the price of worldwide compatibility.
```
utopia-2, line 7: Java `char`s have an allowable range greater than 16 bits (internally it just places 2 chars next to each other) if needed. The code points stop at `0x10FFFF`, allowing _1,114,111_ characters. The book kind of glossed over this though.

mcm/src/mcm_ch2/PrimeNumberCalculator.java
```diff
@@ -0,0 +1,42 @@
...
+	 * @param end
+	 */
+	public static void calculatePrimeNumbers(long start, long end){
+	
+		System.out.println( "The first " + (end++) + " prime numbers are:" );
+		//Uses Java 8's snazzy new lambda expressions to figure out the prime numbers between a range
+		//No idea how efficient these are, just using them to see how they work really
+		LongStream.range(start, end++).forEach(
+			i -> {
+				if(i>=2){
```
utopia-2, line 19: Might be better to check that `start >= 2` outside of the `forEach` instead of on each iteration.

mcm/src/mcm_ch2/PrimeNumberCalculator.java
```diff
@@ -0,0 +1,42 @@
...
+					boolean notprime = LongStream.range(2, i).allMatch(
+							j -> {
+								return (j < i && i%j != 0);
```
utopia-2, line 22: `j` will always be less than `i` (`Stream.range` creates an empty stream when end <= start).

mcm/src/mcm_ch2/PrimeNumberCalculator.java
```diff
@@ -0,0 +1,42 @@
...
+							});
+					System.out.print(notprime ? "Prime: "+i+"\n": "");
```
utopia-2, line 24: The var's called `notprime` but is true when the value _is_ prime.

mcm/src/mcm_ch2/PrimeNumberCalculator.java
```diff
@@ -0,0 +1,42 @@
...
+				}});
```
utopia-2, line 25: It might be neater if instead of printing out whether the value is prime in the lambda body, the stream is _filtered_ based on the predicate that the value is prime, then the filtered stream printed.

mocathe1st, line 25: Yeah, that would make sense. Most of the time on this was spent figuring out how to use the Java 8 lambda expressions, so the overall solution is probably a bit more bloated than it needs to be.

mcm/src/mcm_ch2/PrimeNumberCalculator.java
```diff
@@ -0,0 +1,42 @@
...
+
+	/**
+	 * Main method to run the prime number calculator
+	 * @param args
+	 */
+	public static void main(String[] args) {
+		
+		//Calculate the prime numbers and at the same time
+		//See how hot my Macbook Pro can get
+		calculatePrimeNumbers( 1, 9999999999999L );
```
Spinodal, line 37: probably could have used `Long.MAX_VALUE` for that extra toasty feeling.

PR #20, opened 2014-11-05
Mssw.ch2

Spinodal:
why can I still not push?!?!?!?!


chapter2.md
```diff
@@ -0,0 +1,59 @@
...
+`x = i++` will assign `i` to `x` and then `i` will be incremented.
+
+# 7 #
+```java
+int x = 0;
+int y = 2;
+
+while( x < 10 ){
+	if( x != 0 && (x/y % y) == 0 ){
+		System.out.println( "I think this will print out that x is currently even???" );
```
utopia-2, line 37: You don't need the divide, just `x % y == 0`. e.g. in the above `x = 5` would pass the condition (5/2 = 2; 2 % 2 = 0; _"5 is even???"_).

src/main/java/test/Primer.java
```diff
@@ -0,0 +1,105 @@
...
+			}
+
+			// Print the output
+			System.out.print(toPrime);
+			if (isPrime(toPrime)) {
+				System.out.println(" is prime!");
+			}
+			else {
+				System.out.println(" is not prime");
+			}
```
utopia-2, line 65: The printing behavior and message is basically identical, it might look better extracted to a static method.

src/main/java/test/Primer.java
```diff
@@ -0,0 +1,105 @@
...
+	 * @return Whether the number was prime or not.
+	 */
+	private static boolean isPrime(int isPrime) {
+		for( int x = 2; x < isPrime; x++){
+			if ((isPrime % x) == 0){
+				return false;
+			}
+		}
+		return true;
+	}
```
utopia-2, line 103: The two `isPrime` static methods use the same underlying algorithm, but differ in the types used. It'd be good if there were a generic way to define the types used in this to support both, but I suspect that wouldn't be easy in Java.

Spinodal, line 103: I think the issue would be that `BigInteger` doesn't work with normal maths operators, but i suppose i could have created a wrapper class for `int` to work with.
That might have been a bit of overkill though :P

src/main/java/test/Primer.java
alexdeas, line 65: But we don't know what static methods are yet?!!

utopia-2, line 65: Sure we do, we've unquestioningly used `public static void main( String[] args )`. Also, there are other static methods in this class!

src/main/java/test/Primer.java
utopia-2, line 103: Yeah, I was more complaining that the type hierarchy of `java.lang.Number` is bad enough you can't create general algorithms for different numeric types.

src/main/java/test/Primer.java
mocathe1st, line 65: Wouldn't necessarily have to be a static method, but it would be worth extracting functionality into smaller methods. Maybe for this, have a Primer class, which the main method then instantiates and uses methods from... or just use static methods. Either way would make to code more modular.

src/main/java/test/Primer.java
mocathe1st, line 103: Is there any advantage to using the BigInteger class here over normal Java primitives? Not sure what the win is here with BigInteger?

utopia-2, line 103: It looks like @Spinodal is using `BigInteger` for the user entered values so that whether or not an arbitrarily large number is prime can be determined.

Spinodal, line 103: What @Utopia1138  says is correct, it was so that you can theoretically work out if any number is prime, no matter the size. Just depends on how long you want to wait for it to run :stuck_out_tongue:

PR #21, opened 2014-11-05
Sdt chapter2

SandyThomson:


test/com/selftest/PrimeNumberGeneratorTest.java
```diff
@@ -0,0 +1,25 @@
...
+public class PrimeNumberGeneratorTest {
+
+	@Test
+	public void testPrimesTwoToOneHundred() {
+
+		ArrayList<Integer> primeNumbersForRange = getPrimeNumbersForRange(2,
+				100);
+
+		Assert.assertEquals(
+				Arrays.asList(new Integer[] { 3, 5, 7, 11, 13, 17, 19, 23, 29,
```
SandyThomson, line 20: Oops .. my program doesn't cope with 2. Never mind, I CBA fixing it now.

SandyThomson, line 20: Well, I fixed it in GITHub using the short circuit AND on line 43 of my program. I've not re-ran the test (yet) though.

self_test_answers.txt
```diff
@@ -0,0 +1,29 @@
+Q1: Java strictly specifies the range and behaviour of its primitive types for portability reasons
+
+Q2: Java characters are Unicode (UTF-8), unlike some other languages that just use ASCII.
```
DavidCaldwell, line 3: I think Java uses UTF-16 for strings but char in a numeric type containing a 2 byte unicode codepoint (ie not encoded) - never been totally sure about this though.

PR #22, opened 2014-11-06
Added chapter 2 answers

ardavey:


Primes.java
```diff
@@ -0,0 +1,27 @@
...
+		for ( i = 2; i <= 100; i++ ) {
+			// assume it's prime first of all
+			isPrime = true;
+			
+			// now try to divide by everything between 2 and i/2
+			for ( j = 2; j <= i/2; j++ ) {
+				if ( i % j == 0 ) {
+					// no remainder - we found a divisor. Unset the flag and exit the loop
+					isPrime = false;
+					break;
```
utopia-2, line 19: Alternatively, use a `continue LABEL` statement here and you don't need the `isPrime` boolean or conditional check on the print. e.g.

``` java
PRIMELOOP:
for ( i = 2; i <= 100; i++ ) {
  //... etc.
    if ( i % j == 0 )
      continue PRIMELOOP;
}
```

alexdeas, line 19: Ah, it's continue rather than next. next didn't work when I tried it so I presumed that it was TOTALLY IMPOSSIBLE to break out of a Java loop.

utopia-2, line 19: Yup, and `break` rather than `last`.

Primes.java
```diff
@@ -0,0 +1,27 @@
...
+	public static void main( String[] args ) {
+		long i, j;
+		boolean isPrime;
+		
+		for ( i = 2; i <= 100; i++ ) {
+			// assume it's prime first of all
+			isPrime = true;
+			
+			// now try to divide by everything between 2 and i/2
+			for ( j = 2; j <= i/2; j++ ) {
```
Spinodal, line 15: Not really much of an issue here as you are just working with small numbers, but it would generally be better to save the result of `i/2` to a variable before the loop. That way Java isn't having to re-calculate the value every iteration.

self-test-answers.txt
```diff
@@ -0,0 +1,24 @@
...
+ 3. False.
+ 
+ 4. System.out.println( "One\nTwo\nThree" )
+ 
+ 5. sum is declared and destroyed with every iteration of the for loop, so doesn't actually sum the numbers.
+    It's also locally scoped to the loop so the println() won't work.
+ 
+ 6. Prefix form increments the value before using it. Postfix uses the value and then increments it.
+ 
+ 7. if ( foo != 0 && bar % foo ) { // do stuff }
```
Spinodal, line 14: I'm not sure if you meant to use modulo rather than divide, but as a result of some investigating it does look like `x % 0` produces the same issue!

Primes.java
utopia-2, line 15: @Spinodal `i / 2` in the condition should be optimized by the compiler to the form you describe. The only time this can't be optimized  by the compiler is when you call out to non-inlineable methods, or if you were to change the value of `i` in the loop body.

self-test-answers.txt
utopia-2, line 14: @Spinodal I think the example in the chapter uses modulo for this too? It's just remainder after division, so you'd expect it to exhibit the same behavior.

Primes.java
utopia-2, line 15: ...I just tested this and it didn't optimise! :angry:

``` java
for( int i = 2; i < 100; ++i ) {
  for( int j = 0; j < i / 2; j++ )
    System.out.println(j);
}
```

``` assembly
     0: iconst_2      
     1: istore_1      
     2: goto          30
     5: iconst_0      
     6: istore_2      
     7: goto          20
    10: getstatic     #16    // Field java/lang/System.out:Ljava/io/PrintStream;
    13: iload_2       
    14: invokevirtual #22    // Method java/io/PrintStream.println:(I)V
    17: iinc          2, 1
    20: iload_2       
    21: iload_1       
    22: iconst_2      
    23: idiv          
    24: if_icmplt     10
    27: iinc          1, 1
    30: iload_1       
    31: bipush        100
    33: if_icmplt     5
    36: return        
```

It's possible this kind of thing is sometimes optimised away by the JIT, but I had no such luck:
- Division in conditional:  _113707532_
- Division to a temporary: _22451690_

Note that I removed the `println`, ran the test multiple times and averaged. I also performed the same test with C++ and it _did_ optimise it correctly. (If you still want it in one line `for( int j = 2, lim = i / 2; j <= lim; ++j )`).

PR #24, opened 2014-11-06
Wpb chapt2

WayneBooth:


src/prime.java
```diff
@@ -0,0 +1,19 @@
...
+	public static void main(String args[]) {
+		for (int x = 2; x <= 100; x++) {
+			int isPrime = 1;
+			for (int y = 2; y <= Math.ceil(Math.sqrt(x)); y++) {
+				if (y == x) {
+					break;
+				}
+				if (0 == (x % y)) {
+					isPrime = 0;
+					break;
```
SimonGross: You can use 'continue' to skip to the next step in the outer loop. This would mean you can just put the println in a position where it would be skipped by the continue so you don't need the isPrime variable.

i.e:

``` java
OUTERLOOP:
for ( x = ... )  {
  for  ( y = ... )  {
    if ( ... ) {
     continue OUTERLOOP;
    }
  }
  print...
}
```

src/prime.java
```diff
@@ -0,0 +1,19 @@
...
+				}
+			}
+			if (isPrime == 1) {
```
SimonGross: isPrime would be better off as a `boolean`, because it would allow this statement to become:

``` java
if ( isPrime )
```

src/Sound.java
```diff
@@ -0,0 +1 @@
...
strike whose sound takes 7.2 seconds
to reach you.
 */
class Sound {
	public static void main(String args[]) {
		double dist;
		dist = 7.2 * 1100;
		System.out.println("The lightning is " + dist + " feet away.");
	}
}
```
SimonGross, line 1: You can save a line by performing the calculation for dist on the same line that you declare `dist`

``` java
double dist = 7.2 * 1100;
```

src/LogicalOpTable.java
```diff
@@ -0,0 +1,38 @@
...
+		truth(p, q);
+
+		p = false;
+		q = true;
+		truth(p, q);
+
+		p = false;
+		q = false;
+		truth(p, q);
+	}
```
SimonGross, line 23: As you are not doing anything with the p and q variables except assigning them and passing them into the methods, it would save lines to just remove them entirely and pass the boolean values in directly.

``` java
truth( true, true );
```

Chap2-wpb
```diff
@@ -0,0 +1,54 @@
+Chapter 2 Self Test
+
+	1. 	Why does Java strictly specify the range and behavior of its primitive types?
+        - (book answer) To define the limits of operations on an operand.
+        - (real answer) Becase it does not trust the developer to do the right thing.
+
```
SimonGross, line 6: The actual book answer is to ensure that the primitive types will behave in the same manner on all systems, ensuring portability.

Chap2-wpb
```diff
@@ -0,0 +1,54 @@
...
+		- (real answer) When it doesn't work without casting.
+
+	10. 	Write a program that finds all of the prime numbers between 2 and 100.
+    - Check out "src/prime.java"
+
+	11. 	Does the use of redundant parentheses affect program performance?
+    - Nope
+
+	12. 	Does a block define a scope?
+    - Partially
```
SimonGross, line 53: The answer the book was looking for was 'yes', but I'll accept this if you expand on it.

Chap2-wpb
```diff
@@ -0,0 +1,54 @@
...
+
+	3. 	A boolean value can have any value you like because any non-zero value is true. True or False?	
+        - False
+
+	4. 	Given this output,
+        	One
+        	Two
+        	Three
+	  	using a single string, show the println( ) statement that produced it.
+        - println( "\t\t\tOne\n"
```
alexdeas, line 19: Where did all the tabs come from?

Chap2-wpb
```diff
@@ -0,0 +1,54 @@
...
+                  +"\t\t\tTwo\n"
+                  +"\t\t\tThree\n" );
+
+	5. 	What is wrong with this fragment?
+      for(i = 0; i < 10; i++) {
+        int sum;
+        sum = sum + i;
+      }
+      System.out.println("Sum is: " + sum);
+        - (book answer) 'i' is not defined. 'sum' has not been initialised.
```
alexdeas, line 29: There was one more apparently, which I didn't notice until after.

Chap2-wpb
```diff
@@ -0,0 +1,54 @@
...
+                  +"\t\t\tThree\n" );
+
+	5. 	What is wrong with this fragment?
+      for(i = 0; i < 10; i++) {
+        int sum;
+        sum = sum + i;
+      }
+      System.out.println("Sum is: " + sum);
+        - (book answer) 'i' is not defined. 'sum' has not been initialised. 
+        - (real answer) Someone tried to use Java for such a program.
```
alexdeas, line 30: Good answer

src/LogicalOpTable.java
alexdeas, line 23: Makes it a little clearer for the demo though.

src/prime.java
```diff
@@ -0,0 +1,19 @@
+public class prime {
+	public static void main(String args[]) {
+		for (int x = 2; x <= 100; x++) {
+			int isPrime = 1;
+			for (int y = 2; y <= Math.ceil(Math.sqrt(x)); y++) {
```
alexdeas: Nifty, going as far as the square root is a bit more efficient than going as far as y/2. But who remembers maths anyway?

SimonGross: Using the sqrt does lead to requiring the y == x ( for the case of x = y = 2 ) below. I'm not sure if there's a clean way to use sqrt without that though.

utopia-2: Using `Math.sqrt` for small values of `x`, such as `x < 100` will likely be less efficient than `x / 2`. (See [SO sqrt implementation](http://stackoverflow.com/a/825249) answer). @SimonGross, he doesn't need to worry about the case of `x = 2`, because `isPrime` defaults to true (and `x = 2` doesn't enter the loop body).

src/prime.java
WayneBooth: I'm not sure I understand this, can we talk about it tomorrow?

src/prime.java
WayneBooth: Absolutely agree.

src/prime.java
SimonGross: Sure. It's late enough in the day that any further comments would probably be easier to address in the meeting.

src/Sound.java
WayneBooth, line 1: Agreed.

Chap2-wpb
WayneBooth, line 6: That is a point that I missed (although I do recognise that benefit)

Chap2-wpb
WayneBooth, line 53: Partially, in that the block defines scope, but was I was getting at is - so does the class behavious

Chap2-wpb
WayneBooth, line 19: Artistic licence?

Chap2-wpb
WayneBooth, line 29: :)

src/prime.java
WayneBooth: Agree with @Utopia1138 about efficiency at low values, but Meh - I like the gratuitous use of Math! 
However, consider this story:
The reason I knew about the sqrt thing is I once set a competition where I asked for people to find the first  (I don't know... 1000?) primes, and to do it in the fastest program possible. One quy set an array containing all the primes (up to whatever) and then just iterated over it!

src/prime.java
WayneBooth: Updated now

src/prime.java
WayneBooth: Actually, I got what you are saying. I've updated accordingly - and I like it :)

PR #26, opened 2014-11-06
Completing exercises for Chapter 2.

NormanLittle:


SelfTest/src/com/trythis/FindPrimeNumbers.java
```diff
@@ -0,0 +1,14 @@
+package com.trythis;
+
+import java.util.stream.IntStream;
+
+public class FindPrimeNumbers {
+	
+	public static void main( String[] args ) {
+		IntStream.range(2, 100).filter( i -> {
+			return IntStream.rangeClosed(2, (int)Math.sqrt( i ) )
+							.noneMatch( divisor -> i % divisor == 0 );
```
utopia-2, line 10: You don't need an explicit return or block in a lambda expression (though fair enough, this might be more readable to Java devs unfamiliar with Java 8). :+1: for `noneMatch` use.

SelfTest/src/com/trythis/LogicalOpTable.java
```diff
@@ -0,0 +1,41 @@
...
+												{ false, true  }, 
+												{ false, false } };
+	
+	private static BiFunction<Boolean, Boolean, Stream<Boolean>> getValues = (p,q) -> {
+		return Arrays.stream( new Boolean[] { p, q, (p&q), (p|q), (p^q), (!p) } );
+	};
+	
+	private static BiFunction<Boolean, Boolean, Stream<Integer>> getValuesAsInteger = (p,q) -> {
+		return getValues.apply(p, q).mapToInt( bool -> bool ? 1 : 0 ).boxed();
+	};
```
utopia-2: The Java 8 functional interfaces include `#compose` and `#andThen` for composing functions like this. You can:

``` java
private static BiFunction<Boolean, Boolean, Stream<Integer>> getValuesAsInteger = 
    getValues.andThen( bools -> bools.mapToInt( bool -> bool ? 1 : 0 ).boxed() );
```

SelfTest/src/com/trythis/FindPrimeNumbers.java
mocathe1st, line 10: Like the use of noneMatch - makes for a really clean solution

PR #29, opened 2014-11-11
Red Chapter 3 answers

utopia-2:
Chapter 3 exercises/answers.

### DIFF
--- a/answers.md
+++ b/answers.md
@@ -1,0 +1,121 @@
+## Java Training - Chapter 3 Self Test
+
+  1) _Write a program that reads characters from the keyboard until a period is received.
+     Have the program count the number of spaces. Report the total at the end of the program._
+
+	See `SpaceCounter.java`.
+	
+  2) _Show the general form of the *if-else-if* ladder._
+  
+```java
+if ( condition )
+  statement;
+else if ( condition )
+  statement;
+else
+  statement;
+``` 
+
+  3) _Given_
+```java
+  if(x < 10)
+    if (y > 100) {
+      if(!done) x = z;
+      else y = z;
+    }
+  else System.out.println("error"); // what if?
+```  
+    _to what if does the last *else* associate?_
+    
+    The last else is associated with `if(y > 100)`).
+    
+  4) _Show the *for* statement for a loop that counts from 1000 to 0 by -2._
+   
+```java
+for ( int i = 1000; i > 0; i -= 2 );
+```
+
+  5) _Is the following fragment valid?_
+
+```java
+for(int i = 0; i < num; i++)
+  sum += i;
+  
+count = i;
+```
+
+    No, `i` is defined in the scope of the `for` loop.
+    
+  6) _Explain what *break* does. Be sure to explain both of its forms._
+  
+    The unlabeled form of `break` exits the current block, the labeled form can be used to exit to a labelled 
+    outer block when blocks are arbitrarily nested (as long as the block encloses the block `break` is executed from). 
+    
+  7) _In the following fragment, after the *break* statement executes, what is displayed?_
+  
+```java
+  for(i = 0; i < 10; i++) {
+    while(running) {
+      if(x<y) break;
+      // ...
+    }
+    System.out.println("after while");
+  }
+  System.out.println("After for");
+```
+
+```
+after while
+```
+
+  8) _What does the following fragment print?_
+  
+```java
+for(int i = 0; i<10; i++) {
+  System.out.print(i + " ");
+  if((i%2) == 0) continue;
+  System.out.println();
+}
+```
+
+```
+0 1 
+2 3
+4 5
+6 7
+8 9
+```
+
+  9) _The iteration expression in a *for* loop need not always alter the loop control
+    variable by a fixed amount. Instead, the loop control variable can change in any
+    arbitrary way. Using this concept, write a program that uses a *for* loop to
+    generate progression 1, 2, 4, 8, 16, 32, and so on._
+    
+    And so on? As in without end? Alright. `InfiniteBase2Progression.java`. A bounded
+    solution using a boring *for* loop might look like:
+ 
+ ```java
+for( int i = 1; i > 0; i <<= 1 )
+  System.out.println(i);
+ ```
+ 
+  10) _The ASCII lowercase letters are separated from the uppercase letters by 32. Thus,
+     to convert a lowercase letter to uppercase, subtract 32 from it. Use this information
+     to write a program that reads characters from the keyboard. Have it convert all
+     lowercase letters to uppercase, and all uppercase letters to lowercase, displaying
+     the result. Make no changes to any other character. Have the program stop when the
+     user enters a period. At the end, have the program display the number of case changes
+     that have taken place._
+     
+     See `CaseSwitcher.java`.
+     
+  11) _What is an infinite loop?_
+  
+    A loop with either literally no exit condition (short of killing the program or the power)
+    or a user controlled exit condition (like a game loop that will only exit when the user
+    selects to quit the game) or with a non-deterministic exit condition.
+    
+  12) _When using *break* with a label, must the label be on a block that contains the *break*?_
+  
+    Yes, the labelled block must contain the break, otherwise it's just a filthy `goto`.
+    
--- a/answers.md
+++ b/answers.md
@@ -3,7 +3,7 @@
   1) _Write a program that reads characters from the keyboard until a period is received.
      Have the program count the number of spaces. Report the total at the end of the program._
 
-	See `SpaceCounter.java`.
+See `SpaceCounter.java`.
 	
   2) _Show the general form of the *if-else-if* ladder._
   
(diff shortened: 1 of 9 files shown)


src/org/java/training/SpaceCounter.java
```diff
@@ -0,0 +1,20 @@
...
+ * java.util.stream.Stream.generate with an infinite character reader predicated
+ * on a space was too much of a hassle. 
+ *
+ */
+public class SpaceCounter {
+
+	public static void main(String[] args) throws IOException {
+		int count = 0;
+		for ( char c = 0; c != '.'; c = (char) System.in.read(), count = c == ' ' ? ++count : count );
+		System.out.printf( "Found %d spaces in the supplied input.", count );
```
NormanLittle, line 17: :+1: for printf - how have I not known about this until now?

src/org/java/training/CaseSwitcher.java
```diff
@@ -0,0 +1,23 @@
...
+ */
+public class CaseSwitcher {
+
+	public static void main(String[] args) throws IOException {
+		int caseChanges = 0;
+		for ( int ch, out; ( ch = System.in.read() ) != '.' ; ) {
+			out = ch <= 'Z' && ch >= 'A' ? ch + 32 : ch;
+			out = ch <= 'z' && ch >= 'a' ? ch - 32 : out;
+
+			caseChanges = out == ch ? caseChanges : ++caseChanges;
```
NormanLittle, line 16: :+1: I like this way of counting the case changes.

src/org/java/training/CaseSwitcher.java
```diff
@@ -0,0 +1,23 @@
...
+/**
+ * Chapter 3 self test #10.
+ */
+public class CaseSwitcher {
+
+	public static void main(String[] args) throws IOException {
+		int caseChanges = 0;
+		for ( int ch, out; ( ch = System.in.read() ) != '.' ; ) {
+			out = ch <= 'Z' && ch >= 'A' ? ch + 32 : ch;
+			out = ch <= 'z' && ch >= 'a' ? ch - 32 : out;
```
NormanLittle, line 14: You could have used the _Character.isUpperCase_ and _Character.isLowerCase_ methods which **may** have made it slightly more readable (although that is open to debate).

Also, isn't the second line (checking for lower case characters) executed needlessly in some cases (e.g. when uppercase characters are entered and have already been converted)?

utopia-2, line 14: Use of these methods would be better for production code, but for the sake of the question, it feels like it'd be delegating the solution to the Java libraries.

The second line is indeed needlessly executed when the first one has already converted it, but for the sake of brevity I wanted to leave this as ternary statements rather than an `if-else-if` block. This could be expressed as a nested ternary, but the result might be considered hard to read:

``` java
out = ch <= 'Z' && ch >= 'A' ? ch + 32 :
      ch <= 'z' && ch >= 'a' ? ch - 32 : ch;
```

_(actually, with that indentation, I like the nested ternary solution more, but I assumed someone would complain :laughing:)_

NormanLittle, line 14: Agreed and I did debate (albeit with myself) whether using the _Character_ libraries was really answering the question.  

:+1: I quite like the nested ternary solution too and to be honest it doesn't seem any harder to read (in my opinion).

answers.md
```diff
@@ -0,0 +1,122 @@
...
+     
+  11) _What is an infinite loop?_
+  
+A loop with either literally no exit condition (short of killing the program or the power)
+or a user controlled exit condition (like a game loop that will only exit when the user
+selects to quit the game) or with a non-deterministic exit condition.
+    
+  12) _When using *break* with a label, must the label be on a block that contains the *break*?_
+  
+Yes, the labelled block must contain the break, otherwise it's just a filthy `goto`.
```
mocathe1st, line 121: Good answer, does Java allow any kind of goto? Can't see why it would when its supposed to be all about OO... but not sure

utopia-2, line 121: It reserves the `goto` keyword, but it doesn't do anything.

src/org/java/training/CaseSwitcher.java
mocathe1st, line 14: I'd agree that using the Character library is the best solution for this in "real life" but for the purposes of the exercise this approach is right... I also considered this when doing mine...

The ternary statement is pretty cool. I wonder though, from an efficiency point of view whether an if-else-if block would be more appropriate even if the code's a bit more messy? Thinking when compiled into byte code.

src/org/java/training/CaseSwitcher.java
```diff
@@ -0,0 +1,23 @@
...
+
+			caseChanges = out == ch ? caseChanges : ++caseChanges;
+			System.out.print( (char)out );
+		}
+
+		System.out.printf("%nThis required %d case changes%n", caseChanges);
```
mocathe1st, line 20: Didn't realise Java let you do printf, awesome!

src/org/java/training/SpaceCounter.java
```diff
@@ -0,0 +1,20 @@
...
+ * Chapter 3 self test #1. Opted for a minimal approach because defining
+ * java.util.stream.Stream.generate with an infinite character reader predicated
+ * on a space was too much of a hassle. 
+ *
+ */
+public class SpaceCounter {
+
+	public static void main(String[] args) throws IOException {
+		int count = 0;
+		for ( char c = 0; c != '.'; c = (char) System.in.read(), count = c == ' ' ? ++count : count );
```
mocathe1st, line 16: Nice for statement!

src/org/java/training/CaseSwitcher.java
utopia-2, line 14: Given that the wait on the user filling out the buffer for input would be enormous compared to the cost of a few extra conditions, I wasn't overly concerned with efficiency, but what if we _were_ heavily resource constrained?

**tl;dr** there's not a lot in it.

Here's how different options compare (compiled with OpenJDK 1.8.0):

``` java
out = ch <= 'Z' && ch >= 'A' ? ch + 32 : ch;
out = ch <= 'z' && ch >= 'a' ? ch - 32 : out;
```

``` assembly
       5: iload_2       
       6: bipush        90
       8: if_icmpgt     24
      11: iload_2       
      12: bipush        65
      14: if_icmplt     24
      17: iload_2       
      18: bipush        32
      20: iadd          
      21: goto          25
      24: iload_2       
      25: istore_3      
      26: iload_2       
      27: bipush        122
      29: if_icmpgt     45
      32: iload_2       
      33: bipush        97
      35: if_icmplt     45
      38: iload_2       
      39: bipush        32
      41: isub          
      42: goto          46
      45: iload_3       
      46: istore_3      
      47: iload_3       
      48: iload_2       
```

Next up regular `if-elsif-else`:

``` java
if ( ch <= 'Z' && ch >= 'A' )
  out = ch + 32;
else if ( ch <= 'z' && ch >= 'a' )
  out = ch - 32;
else
  out = ch;
```

``` assembly
       5: iload_2       
       6: bipush        90
       8: if_icmpgt     25
      11: iload_2       
      12: bipush        65
      14: if_icmplt     25
      17: iload_2       
      18: bipush        32
      20: iadd          
      21: istore_3      
      22: goto          47
      25: iload_2       
      26: bipush        122
      28: if_icmpgt     45
      31: iload_2       
      32: bipush        97
      34: if_icmplt     45
      37: iload_2       
      38: bipush        32
      40: isub          
      41: istore_3      
      42: goto          47
      45: iload_2       
      46: istore_3      
      47: iload_3       
      48: iload_2       
```

Same instruction count, but there are some interesting things to note:
    \* The `if-elsif-else` version skips straight to 47, whilst the ternary version has to go to 25 and execute that set of conditions even if the value is already case-switched.
    \* The ternary version doesn't bother to store the value to the output register each branch, but does so once after executing either branch. This can allow ternary statements to take up less instructions than the `if-elsif-else` equivalent.

Finally, my preferred nested ternary solution:

``` java
out = ch <= 'Z' && ch >= 'A' ? ch + 32 :
      ch <= 'z' && ch >= 'a' ? ch - 32 : ch;
```

``` assembly
       5: iload_2       
       6: bipush        90
       8: if_icmpgt     24
      11: iload_2       
      12: bipush        65
      14: if_icmplt     24
      17: iload_2       
      18: bipush        32
      20: iadd          
      21: goto          44
      24: iload_2       
      25: bipush        122
      27: if_icmpgt     43
      30: iload_2       
      31: bipush        97
      33: if_icmplt     43
      36: iload_2       
      37: bipush        32
      39: isub          
      40: goto          44
      43: iload_2       
      44: istore_3      
      45: iload_3       
      46: iload_2       
```

As predicted, we cut down on 2 instructions due to the compiler knowing in advance that the ternary will always assign to the output register - in terms of execution, this is identical to the `if-elsif-else` version, but there are measurable advantages to reducing the instruction count (related to pipelining, and the number of instructions the CPU can fit into the instruction cache before a flush is needed), so ternary's may be more performant.

PR #30, opened 2014-11-11
Completing exercises for Chapter 3

NormanLittle:


src/com/selftest/Convert.java
```diff
@@ -0,0 +1,28 @@
...
+	private static final int TO_LOWER = 32;
+		
+	public static void main( String[] args ) throws IOException {
+		int count = 0;
+		
+		char c;
+		while ( ( c = (char)System.in.read() ) != PERIOD ) {
+			
+			// convert characters within the [A-Za-z] range.
+			if ( c >= 'A' && c <= 'z' ) {
```
codavex: Upper case and lower case letters aren't next to each other in the ascii table (otherwise the constant would be 26), so this will try to convert the 6 characters that separate them as well

NormanLittle: Thanks @codavex .  Tried it last night and confirmed that was the case - guess I should have read the question a bit more thoroughly!

answers.md
```diff
@@ -0,0 +1,68 @@
...
+```java
+for ( int i = 1; i <= 128; i *= 2 )
+	System.out.println( i );
+```
+
+10)  See _Convert.java_.
+
+11)  An infinite loop is a loop without a _valid_ terminating condition.
+
+12)  Yes, but it does not need to be the immediately enclosing block.
```
Spinodal, line 68: all looks good to me!

bin/.gitignore
```diff
@@ -0,0 +1 @@
+/com/
```
Spinodal, line 1: you have a gitignore file and all you added was /com/ ??? 
:stuck_out_tongue:

src/com/selftest/Convert.java
```diff
@@ -0,0 +1,28 @@
+package com.selftest;
+
+import java.io.IOException;
+
+public class Convert {
+
+	private static final char PERIOD = '.';
```
Spinodal, line 7: don't succumb to their propaganda!! its a full stop!

src/com/selftest/Convert.java
```diff
@@ -0,0 +1,28 @@
...
+	public static void main( String[] args ) throws IOException {
+		int count = 0;
+		
+		char c;
+		while ( ( c = (char)System.in.read() ) != PERIOD ) {
+			
+			// convert characters within the [A-Za-z] range.
+			if ( Character.isLetter( c ) ) {
+				count++;
+				c += ( Character.isLowerCase( c ) ? TO_UPPER : TO_LOWER );
```
Spinodal, line 21: this is a nice way of doing it :+1:

bin/.gitignore
NormanLittle, line 1: :) good point, although I think that must have been automatically generated as I have no recollection of creating/adding that.  Honestly!

src/com/selftest/Convert.java
NormanLittle, line 7: It involved less typing! :)

bin/.gitignore
utopia-2, line 1: If you _New Project..._ in eclipse, it automatically detects if the root folder the project's being added to is a git repo, and adds this `.gitignore` for you.

NormanLittle, line 1: Ah, cool.  Next time I'll make more use of it to clean things up a little.

PR #31, opened 2014-11-11
Ajd chapter3

alexdeas:
Chapter 3 done.

### DIFF
--- a/answers/chapter3.md
+++ b/answers/chapter3.md
@@ -3,9 +3,15 @@
   1) _Write a program that reads characters from the keyboard until a period is received.
      Have the program count the number of spaces. Report the total at the end of the program._
 
-	
+See SpaceCounter.java
+
   2) _Show the general form of the *if-else-if* ladder._
   
+```  
+  if (condition) {statement}
+  else if (condition) {statement}
+  else {statement}
+```  
   
   3) _Given_
 ```java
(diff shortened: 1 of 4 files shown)


src/CaseSwitcher.java
```diff
@@ -0,0 +1,37 @@
...
+    System.out.println( "Enter some characters, '.' to end: ");
+    char in;
+    int switchCount = 0;
+    
+    for ( ; ; ) { // infinito loopo
+      in = (char) System.in.read();
+      
+      if ( in == '.' ){
+        break;
+      }
```
SimonGross, line 22: You could also use a do-while loop to achieve the same outcome while making it clear what the loop condition is:

```
do {
...
} while ( in != '.' ) 
```

src/CaseSwitcher.java
```diff
@@ -0,0 +1,37 @@
...
+      else if ( in >= 65 && in <= 90 ) {
+        in += 32;
+        switchCount++;
+      }
+      else if ( in >= 97 && in <=  122 ) {
+        in -= 32;
+        switchCount++;
+      }
```
SimonGross, line 30: This is fine as int and char are interchangeable, but you can also write `in >= 'A' && in <= 'Z'` to make it easy to see exactly what each statement is doing to the char.

answers/chapter3.md
```diff
@@ -0,0 +1,113 @@
...
+```
+
+```
+0 
+1 2 
+3 4
+5 6 
+7 8 
+9 
+```
```
SimonGross, line 87: This actually has the following output:

```
0 1
2 3 
4 5
6 7 
8 9
```

This is because the continue will cause it to skip printing a new line if i is a multiple of 2 (this includes 0). So it goes 0%2 = 0 -> continue, 1%2 = 1 -> print new line, 2%2 = 0 -> continue, etc...

alexdeas, line 87: Yeah, I see I misinterpreted the if statement. Good thing I'm not a compiler.

PR #32, opened 2014-11-12
Wpb chapt3

WayneBooth:


wpb-chapt3
```diff
@@ -0,0 +1,79 @@
...
+and all uppercase letters to lowercase, displaying the result. Make no changes to any other
+character. Have the program stop when the user enters a period. At the end, have the program
+display the number of case changes that have taken place.
+      - [charChange.java]
+
+  11.   What is an infinite loop?
+      - "A loop that does not exit."
+
+  12.   When using break with a label, must the label be on a block that contains the break?
+      - Nope.
```
codavex, line 79: The break must be in the labelled block, but the break may be nested in blocks within the labeled block. The book puts it better than I just did - "The labeled block of code must enclose the break statement, but it does not need to be the immediately enclosing block"

wpb-chapt3
```diff
@@ -0,0 +1,79 @@
...
+  5.  Is the following fragment valid?
+    for(int i = 0; i < num; i++)
+      sum += i;
+    count = i;
+      - Count cannot reference "i". (Assuming "sum" and "num" is defined)
+
+  6.  Explain what break does. Be sure to explain both of its forms.
+      - Move the execution to either:
+          After the current loop
+          or, to the beginning of a named loop
```
codavex, line 36: Execution resumes at the end of a labeled block after a targeted break.

src/charChange.java
```diff
@@ -0,0 +1,23 @@
+import java.io.IOException;
+
+public class charChange {
+  public static void main(String args[]) throws IOException {
+  	char inChar = 0;
+  	while ( true ) {
```
codavex, line 6: This loop should have had a similar exit to readChar.java - it should exit when the user enters a period.

src/readChar.java
```diff
@@ -0,0 +1,21 @@
...
+  	char inChar = 0;
+  	do {
+  		System.out.println("Enter a char and press enter: ");
+  		if( inChar == ' ' ) {
+  			spaceCounter++;
+  		}
+  		char ignore;
+  		inChar = (char) System.in.read();
+  		do {
+  			ignore = (char) System.in.read();
```
codavex, line 15: Not sure I would have done it like this, what if the exit period was in the middle of the ignored characters?

src/charChange.java
```diff
@@ -0,0 +1,23 @@
...
+  		} while(ignore != '\n');
+  		System.out.print("Read: '" + inChar + "', ");
+  		if( inChar >= 'A' && inChar <= 'Z' ) {
+  			inChar += 32;
+  		} 
+  		else if( inChar >= 'a' && inChar <= 'z' ) {
+  			inChar -= 32;
+  		} 
+	  	System.out.println("Out: '" + inChar + "'");
+  	}
```
codavex, line 21: Again, similarly to readChar.java, this should display a count after the exit period is entered.

PR #34, opened 2014-11-12
Andy chapter3

fitzretsil:


src/chapter3/SpaceCounter.java
```diff
@@ -0,0 +1,36 @@
...
+		
+		do {
+			character = (char) System.in.read();
+			if ( character == '.' ) {
+				System.out.println( "Your sentence contained " + spaces + " spaces." );
+				break;
+			} else if ( character == ' ' ) {
+				spaces++;
+			}
+		} while ( character != '\n' );
```
alexdeas, line 32: Didn't need to break out of the loop on newlines. Could have had it exit on a period here.

src/chapter3/MathematicalSeries.java
```diff
@@ -0,0 +1,23 @@
...
+ */
+public class MathematicalSeries {
+
+	/**
+	 * @param args
+	 */
+	public static void main(String[] args) {
+		
+		for ( int i = 1; i < 200; i += i ) {
+			System.out.print( i + " " );
```
alexdeas, line 18: No trailing newline will mess with your terminal after. So inelegant.

src/chapter3/CaseChange.java
```diff
@@ -0,0 +1,45 @@
...
+				changes++;
+				character -= 32;
+			}
+			else if ( character == '.' ) {
+				System.out.println( "." );
+				break;
+			}
+			
+			System.out.print( character );
+		} while ( character != '\n' );
```
alexdeas, line 40: Newline didn't need to be a loop terminator? Why must you discriminate against it?

answers/chapter3.md
```diff
@@ -0,0 +1,119 @@
...
+    if (y > 100) {
+      if(!done) x = z;
+      else y = z;
+    }
+  else System.out.println("error"); // what if?
+```  
+
+_to what if does the last *else* associate?_
+
+  Ans) if (!done) x = z;
```
alexdeas, line 34: That wasn't the last else.

PR #35, opened 2014-11-12
Mssw.ch3

Spinodal:
Chapter 3

### DIFF
--- a/answers/chapter3.md
+++ b/answers/chapter3.md
@@ -138,7 +138,7 @@ for( int i = 1; i < 1024; i += i ){
      user enters a period. At the end, have the program display the number of case changes
      that have taken place._
 
-[CharacterInverter.java](../src/main/java/CharacterInverter.java)
+[CharacterInverter.java](../src/main/java/ch3/CharacterInverter.java)
   
   11) _What is an infinite loop?_
   
--- a/answers/chapter3.md
+++ b/answers/chapter3.md
@@ -138,7 +138,7 @@ for( int i = 1; i < 1024; i += i ){
      user enters a period. At the end, have the program display the number of case changes
      that have taken place._
 
-[CharacterInverter.java](../src/main/java/CharacterInverter.java]
+[CharacterInverter.java](../src/main/java/CharacterInverter.java)
   
   11) _What is an infinite loop?_
   
--- a/answers/chapter3.md
+++ b/answers/chapter3.md
@@ -2,10 +2,24 @@
 
   1) _Write a program that reads characters from the keyboard until a period is received.
      Have the program count the number of spaces. Report the total at the end of the program._
+     
+     (Spaaaaaaaaaaace.java)[src/main/java/ch3/Spaaaaaaaaaaace.java]
 
 	
   2) _Show the general form of the *if-else-if* ladder._
   
+  ```java
+  if ( condition ){
+  	//code
+  }
+  else if ( otherCondition ){
+  	//more code
+  }
+  else if ( someOtherCondition ){
+  	//even more code
+ }
+ //etc
+ ```  
   
   3) _Given_
 ```java
@@ -18,9 +32,17 @@
 ```  
 
 _to what if does the last *else* associate?_
+this would associate with the `if (y > 100) {`
     
     
   4) _Show the *for* statement for a loop that counts from 1000 to 0 by -2._
+  ```java
+  int i = 0;
+  for( int i = 1000; i >= 0; i-=2 ){
+  	//not entirely sure what should be "counted"
+  	System.out.println(++i);
+  }
+  ```
    
   5) _Is the following fragment valid?_
 
@@ -30,8 +52,13 @@ for(int i = 0; i < num; i++)
   
 count = i;
 ```
-    
+No:
+  * `num` is not defined or altered. Could result in an infinte loop.
+  * `sum` is not defined.
+  * `count` is not defined. Also `i` is being assigned to count, but `i` is out of scope by this point.
+     
   6) _Explain what *break* does. Be sure to explain both of its forms._
+  breaks out of the inner most loop it is placed within, or take thread of execution back to a given label.
   
     
   7) _In the following fragment, after the *break* statement executes, what is displayed?_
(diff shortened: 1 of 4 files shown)


answers/chapter3.md
```diff
@@ -0,0 +1,151 @@
...
+this would associate with the `if (y > 100) {`
+    
+    
+  4) _Show the *for* statement for a loop that counts from 1000 to 0 by -2._
+  
+  ```java
+  int i = 0;
+  for( int i = 1000; i >= 0; i-=2 ){
+  	//not entirely sure what should be "counted"
+  	System.out.println(++i);
```
utopia-2, line 45: Why `++i`? I presume by count the question just means decrement the value until 0 is reached.

answers/chapter3.md
```diff
@@ -0,0 +1,151 @@
...
+for(int i = 0; i < num; i++)
+  sum += i;
+  
+count = i;
+```
+
+No:
+  * `num` is not defined or altered. Could result in an infinte loop.
+  * `sum` is not defined.
+  * `count` is not defined. Also `i` is being assigned to count, but `i` is out of scope by this point.
```
utopia-2, line 61: The question probably left out `num`, `sum` and `count` for brevity - assume this is a fragment of code found in a larger block.

answers/chapter3.md
```diff
@@ -0,0 +1,151 @@
...
+     
+  6) _Explain what *break* does. Be sure to explain both of its forms._
+  
+  Breaks out of the inner most loop it is placed within, or take thread of execution back to a given label.
```
utopia-2, line 65: I guess the key point with the second form is that the label must be of an enclosing block, so you can't do full `GOTO` behaviour.

answers/chapter3.md
```diff
@@ -0,0 +1,151 @@
...
+    while(running) {
+      if(x<y) break;
+      // ...
+    }
+    System.out.println("after while");
+  }
+  System.out.println("After for");
+```
+
+Syntax error, x and y are not defined. If by some miracle they were defined, then the following would display:
```
utopia-2, line 81: Cut the book some slack, they're just showing us a fragment! `x` and `y` will have been defined further up.

answers/chapter3.md
```diff
@@ -0,0 +1,151 @@
...
+after while
+after while
+after while
+after while
+after while
+after while
+after while
+after while
+After for
+</pre>
```
utopia-2, line 94: :+1: Indeed. Most of us just mentioned the output of the first statement that follows.

answers/chapter3.md
```diff
@@ -0,0 +1,151 @@
...
+  if((i%2) == 0) continue;
+  System.out.println();
+}
+```
+
+erm...
+```java
+Exception in thread "main" java.lang.ArithmeticException: / by zero
+	at package.Class.method(File.java:lineNumber)
+```
```
utopia-2, line 111: No, the division is by 2 - `0 / 2` is valid.

answers/chapter3.md
```diff
@@ -0,0 +1,151 @@
...
+if we lived in a magical place where we could divide by 0, then the following would happen:
+<pre>
+42 
+1 
+2 3 
+4 5 
+6 7 
+8 9 </pre>
```
utopia-2, line 119: Why 42? Also, `0 / 2` is `0`, so the first line of output will be `0 1`.

answers/chapter3.md
```diff
@@ -0,0 +1,151 @@
...
+  9) _The iteration expression in a *for* loop need not always alter the loop control
+    variable by a fixed amount. Instead, the loop control variable can change in any
+    arbitrary way. Using this concept, write a program that uses a *for* loop to
+    generate progression 1, 2, 4, 8, 16, 32, and so on._
+    
+```java
+for( int i = 1; i < 1024; i += i ){
+  System.out.println( i );
+}
+```
```
utopia-2, line 131: :+1: Good use of `i += i` rather than the wasteful `i * 2`. The question doesn't define an upper bound, so you can do `for ( int i = 1; i > 0; i <<= 1 )` to print the powers of 2 until integer overflow.

answers/chapter3.md
```diff
@@ -0,0 +1,151 @@
...
+     lowercase letters to uppercase, and all uppercase letters to lowercase, displaying
+     the result. Make no changes to any other character. Have the program stop when the
+     user enters a period. At the end, have the program display the number of case changes
+     that have taken place._
+
+[CharacterInverter.java](../src/main/java/ch3/CharacterInverter.java)
+  
+  11) _What is an infinite loop?_
+  
+ A loop which is infinite.
```
utopia-2, line 145: An infinitely recursive definition? :+1:

src/main/java/ch3/CharacterInverter.java
```diff
@@ -0,0 +1,44 @@
...
+			else if ( input >='A' && input <= 'Z'){
+				input += 32;
+			}
+			newString.append(input);
+		}
+		newString.append('.');
+
+		// consume any additional characters
+		while ((input = (char) System.in.read()) != '\n')
+			;
```
utopia-2, line 39: In this case, you don't really need  to bother consuming them - no more input will be read in this program, so it doesn't matter if the input buffer still has characters.

src/main/java/ch3/Spaaaaaaaaaaace.java
```diff
@@ -0,0 +1,38 @@
...
+		// Keep reading until the end of the line.
+		while ((input = (char) System.in.read()) != '.') {
+			if (input == ' ') {
+				spaceCount++;
+			}
+		}
+
+		// consume any additional characters
+		while ((input = (char) System.in.read()) != '\n')
+			;
```
utopia-2, line 33: You don't need to empty the buffer if you won't use it again. Also, while the book recommends this approach, it's a terribly inefficient means of clearing the input buffer - a better solution would be (though admittedly, this has less guarantees about behaviour):

``` java
System.in.skip( System.in.available() );
```

answers/chapter3.md
Spinodal, line 45: i actually have no idea why i printed that....

answers/chapter3.md
Spinodal, line 61: never! :space_invader:

answers/chapter3.md
Spinodal, line 81: i was feeling really sarcastic when doing these the other night :stuck_out_tongue:

answers/chapter3.md
Spinodal, line 111: yea @mocathe1st pointed this out to me, i think i misread and saw `i%0` for some reason...

PR #36, opened 2014-11-13
Ard chapter3

ardavey:


CaseSwap.java
```diff
@@ -0,0 +1,34 @@
+class CaseSwap {
+	
+	public static void main( String[] args ) throws java.io.IOException {
+		
+		int ignore, input, output;
```
utopia-2, line 5: :+1: Good use of `int` instead of `char`s, given that `char` operations elevate operands to `int` anyway.

CaseSwap.java
```diff
@@ -0,0 +1,34 @@
...
+		int swapCount = 0;
+		
+		do {
+			System.out.println( "Enter a character and press return. '.' to end:" );
+			input = System.in.read();
+			
+			do {
+				ignore = System.in.read();
+			} while ( ignore != '\n' );
```
utopia-2, line 14: I don't think the question was asking for a _single_ character to be conditionally case-swapped:

_...Have it convert _all_ lowercase letters to uppercase, and _all_ uppercase letters to lowercase, displaying the result._

I suspect it wants the user to enter in a sentence, like:

```
This sentence will achieve sentience.
```

And output something like:

```
tHIS SENTENCE WILL ACHIEVE SENTIENCE.
There were 32 case conversions. Goodbye!
```

CaseSwap.java
```diff
@@ -0,0 +1,34 @@
...
+			
+			if( 65 <= input && input <= 90 ) {
+				output = input + 32; 
+				swapCount++;
+			}
+			else if( 97 <= input && input <= 122 ) {
```
utopia-2, line 20: These conditions would look clearer with `char` constants (it'll still compare as int). `( 'z' >= input && 'a' <= input )`

CaseSwap.java
```diff
@@ -0,0 +1,34 @@
...
+				output = input - 32;
+				swapCount++;
+			}
+			else {
+				output = input;
+			}
```
utopia-2, line 26: Instead of this `else`, why not set `output` to `input` initially, then do e.g. `output -= 32` in the conditions.

CaseSwap.java
```diff
@@ -0,0 +1,34 @@
...
+			
+			System.out.println( "[" + (char) input + "] -> [" + (char) output + "]" );
```
utopia-2, line 28: We haven't encountered it yet, but `printf` might look clearer here (and lets you not bother casting to char):

``` java
System.out.printf( "[%c] -> [%c]%n", input, output );
```

PowersOfTwo.java
```diff
@@ -0,0 +1,8 @@
+class PowersOfTwo {
+	
+	public static void main( String[] args ) {
+		for( int i = 1; i < 1000000; i += i ) {
```
utopia-2, line 4: :+1: Good use of `i += i` rather than `i * 2` as some others have done. As this is doing powers of 2, you can take advantage of base 2: `for ( int i = 1; i > 0; i <<= 1 )` - the greater than zero check lets you print all powers until the `int` overflows.

ReadCharacters.java
```diff
@@ -0,0 +1,25 @@
...
+		char ignore, input;
+		int spaceCount = 0;
+		
+		do {
+			System.out.println( "Enter a character and press return. '.' to end:" );
+			input = (char) System.in.read();
+			
+			do {
+				ignore = (char) System.in.read();
+			} while ( ignore != '\n' );
```
utopia-2, line 14: As with the case swap question, I don't think it wants the user to have to enter a single `char` per line of input.

self-test.txt
```diff
@@ -0,0 +1,42 @@
...
+    }
+    else if( condition ) {
+      statement;
+    }
+    ...
+    else {
+      statement;
+    }
+
+3.  The last else associates with the first if
```
utopia-2, line 18: Not quite, the `else` binds to the last `if` statement in the same block as itself, but the first `if` doesn't define a block, so the `if` in the same block is actually `if ( y > 100 )`. The indentation makes it confusing.

``` java
  if(x < 10)
    if (y > 100) {
      if(!done) x = z;
      else y = z;
    }
  else System.out.println("error"); // what if?
```

self-test.txt
```diff
@@ -0,0 +1,42 @@
...
+    2 3 
+    4 5 
+    6 7 
+    8 9 
+
+9.  See PowersOfTwo.java
+
+10. See CaseSwap.java
+
+11. A loop whose condition will always be true - it will never naturally end.
```
utopia-2, line 40: Can it _supernaturally_ end? :grinning:

CaseSwap.java
ardavey, line 14: Yeah, you appear to be correct... I'm going to attribute this to sleepy interpretation of the question and failing to engage common sense!

CaseSwap.java
alexdeas, line 28: Spoilers

utopia-2, line 28: @alexdeas I doubt the book will actually cover it, as `printf` doesn't appear to be that popular amongst Java-people. It's the same as the `printf` you know and love from the C family of languages.

self-test.txt
ardavey, line 18: Wow, I totally didn't notice the lack of braces here!

self-test.txt
ardavey, line 40: Yes. :smile:

self-test.txt
codavex, line 18: Too much Python recently?

PR #39, opened 2014-11-13
Chapter 3 answers

GrahamDavidson:


src/Answers.md
```diff
@@ -0,0 +1,17 @@
...
+Q7: "after while" will be displayed at least once followed by "After for"
+Q8: 0 1
+    2 3
+    4 5
+    6 7
+    8 9    
+ Q9: See code.
+ Q10: See code.   
+ Q11: An infinite for loop runs indefinitely for (;;){}
+ Q12: yes the break must be enclosed by the labelled block.
```
NormanLittle, line 17: It is worth mentioning here that it doesn't need to be the immediately enclosing block.

src/Answers.md
```diff
@@ -0,0 +1,17 @@
...
+    break <label> exits/ends the loop with the corresponding label.
+Q7: "after while" will be displayed at least once followed by "After for"
+Q8: 0 1
+    2 3
+    4 5
+    6 7
+    8 9    
+ Q9: See code.
+ Q10: See code.   
+ Q11: An infinite for loop runs indefinitely for (;;){}
```
NormanLittle, line 16: I would have preferred to see you say _why_ the loop runs indefinitely.  Granted, you have shown an example loop but it would be clearer to mention specifically that no valid terminating condition is supplied in an infinite loop.

src/question1/KeyboardReader.java
```diff
@@ -0,0 +1,18 @@
+package question1;
+
+import java.io.IOException;
+
+public class KeyboardReader {
+	public static void main( String[] args ) throws IOException {
+		char key = 'a';
```
NormanLittle, line 7: Initialising the _key_ variable outside of the loop to a random character value seems a bit needless.  Instead, you could have the value set in the loop condition itself:
E.g.

``` java
char key;
while ( ( key = (char)System.in.read() ) != ' ' ) {
...
}
```

src/question1/KeyboardReader.java
```diff
@@ -0,0 +1,18 @@
...
+		int count = 0;
+
+		System.out.println( "Enter some text:" );
+
+		while ( key != '.' ) {
+			key = (char) System.in.read();
+			if ( key == ' ' ) count++;
+		}
+		System.out.println( "There were a total of " + count + " spaces before the first '.' in the input text." );
```
NormanLittle, line 16: I learned this today myself (thanks @Utopia1138) but there is a _System.out.printf_ that can be used to format the output. 
E.g.

``` java
System.out.printf( "There were a total of %d spaces before the first '.' in the input text.", count );
```

src/question1/KeyboardReaderMinimal.java
```diff
@@ -0,0 +1,16 @@
...
+
+import java.io.IOException;
+
+public class KeyboardReaderMinimal {
+	public static void main( String[] args ) throws IOException {
+		int count = 0;
+
+		System.out.println( "Enter some text:" );
+
+		for ( char key = 'a' ; key != '.' ; key = (char) System.in.read() ) {
```
NormanLittle, line 11: :+1: Nice.  Also, you could update your for loop slightly to do this in one line:

``` java
for ( char key = 'a' ; key != '.' ; key = (char) System.in.read(), count = key == ' ' ? ++count : count ); 
```

src/question10/ChangeCase.java
```diff
@@ -0,0 +1,27 @@
+package question10;
+
+import java.io.IOException;
+
+public class ChangeCase {
+
+	public static void main( String[] args ) throws IOException {
+		char key = 'a';
```
NormanLittle, line 8: Same comment as in previous example regarding the initialisation of the _key_ variable etc.

src/question10/ChangeCase.java
```diff
@@ -0,0 +1,27 @@
...
+
+		System.out.println( "Enter some text:" );
+
+		while ( key != '.' ) {
+			key = (char) System.in.read();
+			if ( (key >= 'a') & (key <= 'z') ) {
+				key -= 32;
+				count++;
+			}
+			else if ( (key >= 'A') & (key <= 'z') ) {
```
NormanLittle, line 19: Oops - typo checking for lower case 'z' rather than upper case 'Z' here.  It means you would accidently _convert_ additional characters between the two sets of letters.

src/question2/IfElseLadder.java
```diff
@@ -0,0 +1,11 @@
+package question2;
+
+public class IfElseLadder {
+	public static void main( String[] args ) {
+		int a = 1, b = 2;
+
+		if      ( a > b ) System.out.println( "a is greater than b" );
+		else if ( a < b )	System.out.println( "a is less than b"    );
```
NormanLittle, line 8: I presume your formatting went a little awry here?

PR #40, opened 2014-11-13
Sdt chapter3

SandyThomson:


answers/chapter3.md
```diff
@@ -0,0 +1,117 @@
...
+  5) _Is the following fragment valid?_
+
+```java
+for(int i = 0; i < num; i++)
+  sum += i;
+  
+count = i;
+```
+
+No. `i` is used outwith the scope of the for loop. Also, I would be careful with the for loop ... if `i` starts off greatter than num, then you have an infinite loop scenario.
```
codavex, line 50: good catch on the value of num

answers/chapter3.md
```diff
@@ -0,0 +1,117 @@
...
+}
+```
+
+```
+01
+23
+45
+67
+89
+```
```
codavex, line 88: Missed the spaces

src/com/selftest/CharacterCaseChanger.java
```diff
@@ -0,0 +1,48 @@
...
+					transformationCount++;
+				} else if (Character.isLowerCase(c)) {
+					c = Character.toUpperCase(c);
+					transformationCount++;
+				}
+
+				System.out.print(c);
+
+				if (c == '.') {
+					break LINELOOP;
```
codavex, line 34: Nice use of break and a label, don't think I've seen anyone else use them this week, despite them being in the chapter.

src/com/selftest/SpaceCounter.java
```diff
@@ -0,0 +1,31 @@
...
+		System.out
+				.println("Please enter some characters followed by a line break. Type '.' anywhere in the line to stop");
+
+		int spaceCount = 0;
+
+		/*
+		 * Note that system.in.read doesn't really cope with encoding. If using
+		 * Unicode, we'd want to define the character set /I think/
+		 */
+		int byteChar = ' '; // Initially set this to something other than '.'
```
codavex, line 18: a do-while takes away the need for initialization here

src/com/selftest/CharacterCaseChanger.java
```diff
@@ -0,0 +1,48 @@
...
+
+import java.io.IOException;
+import java.util.Scanner;
+
+public class CharacterCaseChanger {
+
+	public static void main(String... args) throws IOException {
+
+		System.out
+				.println("Please enter some characters followed by a line break. Type '.' anywhere in the line to stop");
```
codavex, line 11: looks 'broken' broken up like this, think I'd have put just the parameter to println on the new line

answers/chapter3.md
utopia-2, line 50: If `i` starts off greater than `num` won't it just not execute the loop (the condition `i < num` will fail)?

codavex, line 50: Ach, yeah, (mustn't just read answer, must re-read question)

PR #45, opened 2014-11-19
Chapter 4 answers

alexdeas:
Chapter 4 answers

### DIFF
--- a/answers/chapter4.md
+++ b/answers/chapter4.md
@@ -1,0 +1,27 @@
+1. A class is an abstract definition of an object. An object is an instance of a class.
+
+2. A class is defined with the ``class`` keyword.
+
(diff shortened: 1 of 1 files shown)


answers/chapter4.md
```diff
@@ -0,0 +1,27 @@
+1. A class is an abstract definition of an object. An object is an instance of a class.
+
+2. A class is defined with the ``class`` keyword.
+
+3. Objects have their own copies of any class attributes, as long as they're not static.
+
+4. ``MyCounter counter; counter = new MyCounter();``
+``MyCounter counter = new MyCounter();``
```
SimonGross, line 8: I think this was an ambiguous question, so it's not really a mark against you, but the book was only looking for the first line here (as it is made up of two statements).

PR #46, opened 2014-11-19
chapter 4 answers

Spinodal:


answers.md
```diff
@@ -0,0 +1,39 @@
+# Chapter 4 Answers
+
+1) A class is the abstract form of an object. An object is a concrete form of a class.
```
utopia-2, line 3: This terminology is a little misleading, given that `abstract` usually refers to an incomplete implementation of a contract, and `concrete` often refers to a complete implementation of a contract (where contract refers to the externally accessible operations/members of a type).

An object is an instantiation of a class, which is to say an allocated chunk of memory containing the instance member variables and a reference to the methods (for dynamic dispatch) _defined_ by the class.

answers.md
```diff
@@ -0,0 +1,39 @@
...
+
+2) 
+```java
+class ClassName {
+
+}
+```
+
+3) Instance variables.
```
utopia-2, line 12: :+1: It's also notable that each instance will get a reference to a structure containing _class_ data used for such things as dynamic dispatch and reflection see [Object Representation](http://www.artima.com/insidejvm/ed2/jvm6.html), particularly _Figure 5-6_.

answers.md
```diff
@@ -0,0 +1,39 @@
...
+
+4) 
+```java
+MyCounter counter = new MyCounter();
+```
```
utopia-2, line 17: The question asks for this as two statements for whatever reason.

answers.md
```diff
@@ -0,0 +1,39 @@
...
+``` 
+
+5)
+```java
+	public double myMeth( int a, int b ) {
+	
+	}
+```
+
+6) In order for a method to return from a method with a return value, it must return with a value.
```
utopia-2, line 26: :+1: (unless it doesn't return normally by throwing an exception, but we haven't seen them yet in the book)

answers.md
```diff
@@ -0,0 +1,39 @@
...
+
+7) The constructor has the same name as the class.
+
+8) `new` will create an instance of a class, as an object.
+
+9) Garbage collection will hoover up all objects that have dropped out of scope, and therefore cannot be used anymore.
```
utopia-2, line 32: GC won't necessarily clean up **all** objects that have become unreferred, in simple cases it's more efficient to just leave the memory allocated and let the OS do the clean up for you when the process exits (note that instance finalization isn't _ever_ guaranteed to happen).

answers.md
```diff
@@ -0,0 +1,39 @@
...
+```
+
+6) In order for a method to return from a method with a return value, it must return with a value.
+
+7) The constructor has the same name as the class.
+
+8) `new` will create an instance of a class, as an object.
+
+9) Garbage collection will hoover up all objects that have dropped out of scope, and therefore cannot be used anymore. 
+Nobody knows how garbage collection works. It has a mind of its own, but generally will run when there have been a lot of instance variables hanging around in memory not being used.
```
utopia-2, line 33: :+1: for nobody knows how GC works. :smile:

answers.md
Spinodal, line 17: my bad:

``` java
MyCounter counter;
counter = new MyCounter();
```

PR #47, opened 2014-11-20
Red - Chapter 4 exercises/test answers

utopia-2:
Chapter 4 _Try This_ and self test answers.

### DIFF
--- a/answers.md
+++ b/answers.md
@@ -1,0 +1,105 @@
+## Chapter 4 Self Test
+
+ 1) _What is the difference between a class and an object?_
+ 
+ A class defines the contained types, layout, and operations that can be performed on class 
+ instances (the book uses the term _template_, which is reasonable, but unnecessarily confuses
+ it with the template metaprogramming concept in languages like C++).
+ 
+ An object is an _instance_ of a class (where per-instance state is allocated in
+ memory). It should be noted that in Java classes can also have instantiable state (static
+ members); unlike instances, these are allocated at _class load time_.
+ 
+ 2) _How is a class defined?_
+ 
+ ```java
+ <access modifier|default> <abstract?> class <ClassName> <Generic parameters?> <extends|implements? SomeOtherClass> {
+   //...
+ }
+ ```
+ 
+ For example:
+ 
+ ```java
+ public class Kitteh {}
+ ```
+ 
+ Or for more complex classes:
+ 
+ ```java
+ public abstract class Convoluted<T extends Thing> extends SimplerBase implements Function<T, String>, Callable<T> {}
+ ```
+ 
+ 3) _What does each object have its own copy of?_
+ 
+ All non-static member variables defined in the class (or base class).
+ 
+ 4) _Using two separate statements, show how to declare an object called `counter` of a class called `MyCounter`_
+ 
+ ```java
+ MyCounter counter;
+ counter = new MyCounter(1);
+ ```
+ 
+ 5) _Show how a method called `myMeth()` is declared if it has a return type of `double` and has two `int`
+   parameters called a and b._
+   
+```java
+  public double myMeth( int a, int b ); 
+```
+ 
+ 6) _How must a method return if it returns a value?_
+ 
+ It must explicitly define `return <type>;` statements wherever control would be returned to the caller. The
+ one exception to this is, well, throwing exceptions.
+ 
+ 7) _What name does a constructor have?_
+ 
+ The identifier of the class it belongs to.
+ 
+ 8) _What does `new` do?_
+ 
+ Creates an instance of a type by dynamically allocating memory, then calling the supplied constructor for
+ the type, returning a reference to the newly allocated instance. In some circumstances it'll throw a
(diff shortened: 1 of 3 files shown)


answers.md
```diff
@@ -0,0 +1,105 @@
...
+
+ 1) _What is the difference between a class and an object?_
+ 
+ A class defines the contained types, layout, and operations that can be performed on class 
+ instances (the book uses the term _template_, which is reasonable, but unnecessarily confuses
+ it with the template metaprogramming concept in languages like C++).
+ 
+ An object is an _instance_ of a class (where per-instance state is allocated in
+ memory). It should be noted that in Java classes can also have instantiable state (static
+ members); unlike instances, these are allocated at _class load time_.
```
mocathe1st, line 11: Good answer

answers.md
```diff
@@ -0,0 +1,105 @@
...
+ 
+ 2) _How is a class defined?_
+ 
+ ```java
+ <access modifier|default> <abstract?> class <ClassName> <Generic parameters?> <extends|implements? SomeOtherClass> {
```
mocathe1st, line 16: Nice

answers.md
```diff
@@ -0,0 +1,105 @@
...
+ 
+ Creates an instance of a type by dynamically allocating memory, then calling the supplied constructor for
+ the type, returning a reference to the newly allocated instance. In some circumstances it'll throw a
+ runtime exception (e.g. `OutOfMemoryError`).
+ 
+ 9) _What is garbage collection, and how does it work? What is `finalize()`?_
+ 
+ Garbage collection is a means some languages employ of cleaning up dynamically allocated memory for reuse
+ (languages that don't trust the programmer to properly employ RAII techniques and clean up memory when it's
+ actually no longer needed as opposed to, say, whenever).
```
mocathe1st, line 70: Can programmers always be trusted to remember memory allocation ;)

answers.md
```diff
@@ -0,0 +1,105 @@
...
+ 
+ `finalize` is a method that gets called on an instance when it's eligible for garbage collection but prior
+ to collection. It can be used to clean up resources owned by the instance, or more dangerously to cause
+ the instance to become ineligible for collection (_~back from the dead~_).
+ 
+ It should be noted that whilst Java handles memory automatically, this is not so for various other resources
+ (file handles, database connections, socket connections, basically any other resource with a concept of
+ ownership) - here Java _does_ employ RAII with the `AutoCloseable` interface, and notably the new 
+ `try-with-resources` semantics:
+
```
mocathe1st, line 83: Has this changed somewhat in the newer versions of Java. I seem to remember having read something about file handlers etc being automatically closed in newer Java versions, but could have completely made that up.

mocathe1st, line 83: Yeah, try with resources apparently reduces the need for close statements, although from the way this is implemented, you may as well have a close statement really.
[http://java.dzone.com/articles/java-7-new-feature-%E2%80%93](http://java.dzone.com/articles/java-7-new-feature-%E2%80%93)

Example:

``` java
    File file = new File("input.txt");  

     try (InputStream is = new FileInputStream(file)) {  
        // do something with this input stream  
       // ...  
    }  
     catch (FileNotFoundException ex) {   
    System.err.println("Missing file " + file.getAbsolutePath());  
    }  

//No file.close()
```

answers.md
utopia-2, line 70: Of course :wink: (okay, we totally can't).

My point was more that good use of RAII negates the need to think about cleanup in common cases - that is, accessing resources through a handle that cleans itself up when it (the handle), is out of scope. For example, the following C code causes a memory leak:

``` c
MyBuffer* init_buffer() {
  MyBuffer* buffer = malloc( sizeof(*buffer) );
  /* initialise buffer */
  return buffer;
}

void do_something() {
  MyBuffer* buff = init_buffer();
} /* Oh no, buffer was never freed! */
```

But the following modern C++ code equivalent is memory-safe:

``` c++
unique_ptr<MyBuffer> init_buffer() {
  unique_ptr<MyBuffer> buffer { new MyBuffer };
  // initialise buffer
  return buffer;
}

void do_something() {
  auto buff = init_buffer();
} /* buffer is freed automatically */
```

I guess I wish Java had some mechanism for scope-based deallocation for those cases when you have a clearly defined resource owner.

answers.md
utopia-2, line 83: Yeah, that's what I was getting at - Java employs RAII for resources except memory. In my `AudioInputStream` example below, if a `RuntimeException` is thrown, our file is still closed, because the resource leaving scope causes its `close` method to be called regardless.

This all helps reduce the worst examples of Java boilerplate I've seen:

``` java
InputStream in = null;
try {
  in = new FileInputStream( new File("input.svg") );
  /* do stuff with it... */
} catch ( FileNotFoundException e ) {
  /* log it and cry about it later */
} finally {
  if ( in != null ) {
    try {
      in.close();
    } catch( IOException e ) {
      /* log it and shake fist at Java */
    }
  }
}
```

PR #53, opened 2014-11-26
spg Chapter Five

SimonGross:
Chapter Five answers

### DIFF
--- a/src/com/datacash/training/Bubble.java
+++ b/src/com/datacash/training/Bubble.java
@@ -1,0 +1,80 @@
+
+package com.datacash.training;
+
+import java.util.Arrays;
+import java.util.List;
+
+public class Bubble {
+
+	public static void main( String[] args ) {
+
+		bubbleSortArray();
+
+		bubbleSortList();
+
+	}
+
+	private static void bubbleSortArray() {
+
+		int nums[] = { 99, -10, 100123, 18, -978, 5623, 463, -9, 287, 49 };
+
+		int temp;
+		int size = nums.length;
+
+		// Display original array
+		System.out.println( "Original array is:" );
+		for ( int n = 0 ; n < size ; n++ ) {
+			System.out.print( " " + nums[n] );
+		}
+		System.out.println();
+
+		// bubble sort
+		for ( int a = 1 ; a < size ; a++ ) {
+			for ( int b = size - 1 ; b >= a ; b-- ) {
+				if ( nums[b - 1] > nums[b] ) {
+					temp = nums[b - 1];
+					nums[b - 1] = nums[b];
+					nums[b] = temp;
+				}
+			}
+		}
+
+		// Display sorted array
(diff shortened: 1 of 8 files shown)


src/com/datacash/training/Bubble.java
```diff
@@ -0,0 +1,80 @@
...
+		System.out.println( "Sorted array is:" );
+		for ( int n = 0 ; n < nums.length ; n++ ) {
+			System.out.print( " " + nums[n] );
+		}
+		System.out.println();
+
+	}
+	
+	private static void bubbleSortList() {
+		List<Integer> nums = Arrays.asList( 99, -10, 100123, 18, -978, 5623, 463, -9, 287, 49 );
```
alexdeas, line 52: So is using Lists faster than Arrays or what?

utopia-2, line 52: Due to method overhead, slightly slower, @alexdeas, but `List` has the `stream` method :grinning:.

src/com/datacash/training/Question7.java
```diff
@@ -0,0 +1,27 @@
...
+
+		System.out.print( "Original message: " );
+		System.out.println( msg );
+
+		msg.chars().forEach( ( ch ) -> encMsg.append( (char) (ch ^ key.charAt( msg.indexOf( ch ) % 8 )) ) );
+
+		System.out.print( "Encoded message: " );
+		System.out.println( encMsg );
+
+		encMsg.chars().forEach( ( ch ) -> decMsg.append( (char) (ch ^ key.charAt( encMsg.indexOf( Character.toString( (char) ch ) ) % 8 )) ) );
```
alexdeas, line 21: What's with the turning the key character back into a string? It should be possible to decode with the same operations as encoding.

src/com/datacash/training/chapter5.md
```diff
@@ -0,0 +1,37 @@
...
+
+11) Both these cases will cause an error, as the boundaries of an array are strictly enforced
+
+12) >>>
+
+13) See Question13.java
+
+14) The for loops cannot be converted into for-each loops, as the loop control variables are required to assign values to the area when sorting. It is possible to do this when using Lists as demonstrated in Question 4, although it is not as efficient.
+
+15) As of Java 7, switch statements can be controlled by a String.
```
alexdeas, line 37: Correct answers!

src/com/datacash/training/Question7.java
SimonGross, line 21: For whatever reason `StringBuilder.indexOf` will only accept a String, not a char. In hindsight I could have called `toString` at some point to make sure the same operations were being performed.

src/com/datacash/training/chapter5.md
codavex, line 37: Agreed!

SimonGross, line 37: :tada:

PR #55, opened 2014-11-26
Sdt chapter5

SandyThomson:


answers/chapter5.md
```diff
@@ -0,0 +1,66 @@
+Q1
+
+       double array[] = new double[12];
+       double[] array = { 1, 2, 3, 4, 5, 6, 7, 8, 9, 10.01, 11.02, 12.03 };
+
```
SimonGross, line 5: I think it was asking for two ways to declare an array, not initialize it with values (i.e. like in Question 2). In that case the second line should be:

``` java
double[] array = new double[12];
```

answers/chapter5.md
```diff
@@ -0,0 +1,66 @@
...
+		See Encode(Test).java
+		
+Q8
+
+		No
+
+Q9
+
+		y = (x < 0) ? 10 : 20;
+
```
SimonGross, line 40: You don't technically need the brackets, but they could be useful for readability.

answers/chapter5.md
```diff
@@ -0,0 +1,66 @@
...
+Q12
+
+		>>>
+		
+		It shifts bits right, but zerofills from the left
+		
+Q13
+
+		// TODO
+
```
SimonGross, line 59: Tsk, tsk

src/com/selftest/DoubleAverage.java
```diff
@@ -0,0 +1,23 @@
...
+		int count = 0;
+		double sum = 0;
+
+		for (double val : list) {
+			sum += val;
+			count++;
+		}
+
+		if (count > 0) {
+			return sum / count;
```
SimonGross, line 16: You can achieve the same effect by initializing `count` to `list.length` and not incrementing it.

src/com/selftest/StringArraySorter.java
```diff
@@ -0,0 +1,33 @@
...
+
+		int size = strings.length;
+
+		for (int a = 1; a < size; a++) {
+			for (int b = size - 1; b >= a; b--) {
+
+				// Start off by ignoring case, and then only factor in case if
+				// it might make a difference
+				int compareToResult = strings[b - 1]
+						.compareToIgnoreCase(strings[b]);
```
SimonGross, line 15: :+1: for taking upper/lower case into account. 

The question only asked us to sort the Strings and didn't give criteria for sorting and whether case mattered, so either way is valid imo.

test/com/selftest/EncodeTest.java
```diff
@@ -0,0 +1,54 @@
...
+import org.junit.Test;
+
+public class EncodeTest {
+
+	@Test
+	public void testEncodeDecodeLongerString() throws Exception {
+		
+		Encode encode = new Encode();
+		
+		encode.setKey("boobs");
```
SimonGross, line 13: Very minor, but the question did ask that the key was 8 characters.

answers/chapter5.md
SandyThomson, line 5: Yeah. I just wanted to make things more different.

src/com/selftest/DoubleAverage.java
SandyThomson, line 16: Yeah. I realised my mistake when I started to review other peoples work.

src/com/selftest/StringArraySorter.java
SandyThomson, line 15: +1

test/com/selftest/EncodeTest.java
SandyThomson, line 13: _Totally_ missed that... sigh.

codavex, line 13: booooobs

PR #56, opened 2014-11-26
chapter 5 answers

Spinodal:


src/main/java/Averagenator.java
```diff
@@ -0,0 +1,43 @@
...
+ * @author Mike
+ *
+ */
+public class Averagenator {
+
+	/**
+	 * @param args not used
+	 */
+	public static void main(String[] args) {
+		int[] values = new int[10];
```
NormanLittle, line 18: The question asked for **double** values to be used :)

src/main/java/Encode.java
```diff
@@ -0,0 +1,47 @@
...
+package main.java;
+
+import java.util.Scanner;
+
+/**
+ * @author Mike
+ *
+ */
+public class Encode {
+	public static String KEY = "Blarg Blarg Honk";
```
NormanLittle, line 13: It doesn't really matter but, technically, the question asked for a **8 character String** to be used for the key :)

src/main/java/Encode.java
```diff
@@ -0,0 +1,47 @@
...
+
+	}
+
+	public static String encodeAndDecode(String text, String key) {
+		StringBuilder sb = new StringBuilder();
+		int keyIndex = 0;
+		char[] letters = text.toCharArray();
+		for (char letter : letters) {
+			sb.append((char) (letter ^ key.charAt(keyIndex)));
+			keyIndex = (keyIndex + 1 < key.length()) ? ++keyIndex : 0;
```
NormanLittle, line 41: You could have just included the _modulus_ operation in the line above.  E.g. Something like...

``` java
sb.append((char) (letter ^ key.charAt(keyIndex % key.length())));
```

src/main/java/MinMax.java
```diff
@@ -0,0 +1,31 @@
...
+	 * @param args not used
+	 */
+	public static void main(String[] args) {
+		int[] nums = { 99, -10, 100123, 18, -978, 5623, 463, -9, 287, 49 };
+
+		int min, max;
+
+		min = max = nums[0];
+		for (int num : nums) {
+			if (num < min)
```
NormanLittle, line 22: You could have used the ternary operator here.  E.g.

``` java
min = num < min ? num : min;
```

src/main/java/StringSorter.java
```diff
@@ -0,0 +1,67 @@
...
+
+	/**
+	 * This will sort an array of objects that implement the Comparable
+	 * interface.
+	 * 
+	 * @param toSort
+	 *            An array of objects to sort
+	 */
+	@SafeVarargs
+	public static <T extends Comparable<T>> void sorter(T... toSort) {
```
NormanLittle, line 31: :+1:  for generics!

src/main/java/StringSorter.java
```diff
@@ -0,0 +1,67 @@
...
+
+	/**
+	 * This will loop through an array and print each element separated by a
+	 * space.
+	 * 
+	 * @param toPrint
+	 *            An array object print out
+	 */
+	@SafeVarargs
+	public static <T extends Object> void printer(T... toPrint) {
```
NormanLittle, line 60: Nice.  You could also replace this method by using the stream functionality. E.g.

``` java
System.out.printf("After sorting: %s%n", Arrays.stream( toPrint ).collect(Collectors.joining(" ")));
```

PR #57, opened 2014-11-26
Red chapter 5 exercises

utopia-2:
Chapter 5 exercises, answers and _Try This_ tweaks.

### DIFF
--- a/answers.md
+++ b/answers.md
@@ -19,11 +19,11 @@
   3) _Write a program that uses an array to find the average of 10 **double values**. Use any
      10 values you like._
      
-See `Average.java`.
+See [Average.java](src/org/java/training/Average.java).
 
   4) _Change the sort in Try This 5-1 so that it sorts an array of strings. Demonstrate that it works._
   
-See `StringSort.java`
+See [StringSort.java](src/org/java/training/StringSort.java).
 
   5) _What is the difference between the String methods `indexOf()` and `lastIndexOf()`?_
   
@@ -47,7 +47,7 @@ Or without assigning to a reference:
 
   7) _Expanding on the `Encode` cipher class, modify it so that it uses an eight-character string as the key._
   
-See `BadEncode.java`.
+See [BadEncode.java](src/org/java/training/BadEncode.java).
 
   8) _Can the bitwise operators be applied to the `double` type?_
   
@@ -83,7 +83,7 @@ A bitwise right-shift that doesn't extend the sign bit. For example signed right
 
  13) _Rewrite the `MinMax` class shown earlier in this chapter so that it uses a for-each style `for` loop._
  
- See `MinMax.java`.
+ See [MinMax.java](src/org/java/training/MinMax.java).
  
  14) _Can the `for` loops that perform sorting in the `Bubble` class shown in `Try This 5-1` be converted into
      for-each style loops? If not, why not?_
--- a/answers.md
+++ b/answers.md
@@ -1,0 +1,96 @@
+## Chapter 5 Self Test
+
+  1) _Show two ways to declare a one-dimensional array of 12 **doubles**._
+  
+ ```java
+ double[] vals = new double[12];
+ ```
+ 
+ ```java
+ double vals[] = new double[12];
+ ```
+ 
+  2) _Show how to initialize a one-dimensional array of integers to the values 1 through 5._
+   
+ ```java
+ int[] vals = { 1, 2, 3, 4, 5 };
+ ```
+ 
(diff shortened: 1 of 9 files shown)


src/org/java/training/Average.java
```diff
@@ -0,0 +1,14 @@
+package org.java.training;
+
+public class Average {
+	
+	public static void main(String[] args) {
+		double[] vals = { 24.03, 25_325.34, 22.0, 1.5, -3_994.543, 2_435.33, 100.00, 293.34, 34, 2_385.34 };
```
Spinodal, line 6: `_`'s ftw!!!

src/org/java/training/StringSort.java
```diff
@@ -0,0 +1,133 @@
...
+import java.util.Random;
+import java.util.function.BiConsumer;
+import java.util.function.Supplier;
+import java.util.stream.Collectors;
+
+public class StringSort {
+	
+	public static void main(String[] args) {
+		String[] message = { "This", "is", "an", "array", "of", "strings", "sort", "them", "please.", "I'll", "wait..." };
+		Supplier<String> join = () -> Arrays.stream(message).collect(Collectors.joining(" "));
```
Spinodal, line 16: greatly confused...

answers.md
```diff
@@ -0,0 +1,96 @@
...
+```java
+String like = "I like Java";
+int len = like.length();
+char ch = like.charAt(4);
+```
+
+Or without assigning to a reference:
+
+```java
+"I like Java".charAt(3);
```
Spinodal, line 45: I think it was wanting this way to illustrate you can call a method on a string like this

src/org/java/training/BadEncode.java
```diff
@@ -0,0 +1,20 @@
...
+		StringBuilder enc = new StringBuilder(), dec = new StringBuilder();
+		for(int i = 0, lim = key.length(); i < msg.length(); enc.append((char)( key.charAt(i % lim) ^ msg.charAt(i++) ) ) );
+		
+		System.out.println("Encoded message: " + enc);
+		
+		for(int i = 0, lim = key.length(); i < enc.length(); dec.append((char)( key.charAt(i % lim) ^ enc.charAt(i++) ) ) );
+		
+		System.out.println("Decoded message: " + dec);
+	}
+}
```
Spinodal, line 20: too many one liners!!!

answers.md
```diff
@@ -0,0 +1,96 @@
...
+
+ 13) _Rewrite the `MinMax` class shown earlier in this chapter so that it uses a for-each style `for` loop._
+ 
+ See [MinMax.java](src/org/java/training/MinMax.java).
+ 
+ 14) _Can the `for` loops that perform sorting in the `Bubble` class shown in `Try This 5-1` be converted into
+     for-each style loops? If not, why not?_
+     
+No. Sorting algorithms operate by comparing and swapping values held at different indices in each iteration
+whereas the for-each loop operates on one value at a time.
```
Spinodal, line 92: I think you could technically sort the elements into a new array/list, but that might not be the most efficient...

utopia-2, line 92: You could insert it into a self-ordering structure like a binary tree, but then it's not the same as the for-each loop doing the sorting.

src/org/java/training/BadEncode.java
utopia-2, line 20: The right amount! :sunglasses:

src/org/java/training/StringSort.java
utopia-2, line 16: It's an object that wraps the `message` array to make it easy to later print out the arrays contents with just `join.get()`. `collect` takes a `Collector` that builds up a result from each element of a stream, `Collectors.joining` provides a `Collector` that joins each `String`, placing the delimiter between each.

answers.md
utopia-2, line 45: I didn't think that'd be what it was after, as it didn't really get into string literals as object references in the chapter. Taking a look, this _is_ what it wanted though - ah well, adding both is always a safe bet :smile:.

src/org/java/training/BadEncode.java
utopia-2, line 20: The duplicate behaviour could be reduced, but I wanted to more or less follow the book, with improvements:

``` java
Function<String, Supplier<String>> cipher = s -> {
  int j = 0;
  return () -> msg.chars().map(c -> (char)c ^ s.charAt(j++ % s.length())
                          .map(c -> c.toString()).collect(Collectors.joining());
  };
```

Then later:

``` java
String encoded = cipher.accept(msg).get();
String decoded = cipher.accept(encoded).get();
```

PR #58, opened 2014-11-27
Ard chapter5

ardavey:


self-test.md
```diff
@@ -0,0 +1,66 @@
...
+
+##### _9. Show how this sequence can be rewritten using the ? operator._
+```java
+    if(x < 0) y = 10;
+    else y = 20;
+```
+
+```java
+( x < 0 ) ? y = 10 : y = 20;
+```
```
utopia-2, line 43: That's an interesting way of doing it. Assignment _does_ return the assigned value, so this is allowed, but I'm not sure where it'd be useful compared to `y = x < 0 ? 10 : 20;`.

self-test.md
```diff
@@ -0,0 +1,66 @@
...
+    // ...
+    if(a & b) ...
+```
+The `&` is logical here because the operands are boolean.
+
+##### _11. Is it an error to overrun the end of an array? Is it an error to index an array with a negative value?_
+Yes, and yes.
+
+##### _12. What is the unsigned right-shift operator?_
+`>>>`
```
utopia-2, line 57: I think it was looking for an explanation.

src/selftest/Encode.java
```diff
@@ -0,0 +1,37 @@
...
+		System.out.print("Original message: ");
+		System.out.println(msg);
+
+		// encode the message
+		j = 0;
+		for (int i = 0; i < msg.length(); i++) {
+			encmsg = encmsg + (char) (msg.charAt(i) ^ key.charAt(j++));
+			if (j == key.length()) {
+				j = 0;
+			}
```
utopia-2, line 21: This is where the modulo operator is useful, the if can be replaced by: `j %= key.length();`.

self-test.md
```diff
@@ -0,0 +1,66 @@
...
+No.
+
+##### _9. Show how this sequence can be rewritten using the ? operator._
+```java
+    if(x < 0) y = 10;
+    else y = 20;
+```
+
+```java
+( x < 0 ) ? y = 10 : y = 20;
```
NormanLittle, line 42: I'm pretty sure that won't compile as the ternary should be used with values (rather than other expressions).

``` java
y = x < 0 ? 10 : 20;
```

self-test.md
```diff
@@ -0,0 +1,66 @@
...
+Yes, and yes.
+
+##### _12. What is the unsigned right-shift operator?_
+`>>>`
+
+##### _13. Rewrite the MinMax class shown earlier in this chapter so that it uses a for-each style for loop._
+See [MinMax.java](src/selftest/MinMax.java)
+
+##### _14. Can the for loops that perform sorting in the Bubble class shown in Try This 5-1 be converted into for-each style loops? If not, why not?_
+The outer loop cannot be rewritten as a for-each because the inner loop uses its counter variable.  The inner one cannot be rewritten because its loop variable is needed to index array elements.
```
NormanLittle, line 63: Also, you cannot change the underlying array with the for-each loop.

self-test.md
```diff
@@ -0,0 +1,66 @@
...
+
+##### _15. Can a String control a switch statement?_
+Yes, as of Java 7.
```
NormanLittle, line 66: :+1: for including _when_ this was added to Java.

self-test.md
utopia-2, line 43: Testing this after seeing @NormanLittle 's comment, this won't compile in the form you've shown - to get it to compile the whole expression needs an lvalue, and each assignment needs to be wrapped in parentheses:

``` java
y = x < 0 ? (y = 10) : (y = 20);
```

src/selftest/Encode.java
```diff
@@ -0,0 +1,37 @@
+package selftest;
+
+//Use XOR to encode and decode a message.
+public class Encode {
+	public static void main(String args[]) {
+		String msg = "This is a test";
+		String encmsg = "";
+		String decmsg = "";
+		String key = "foobar";
```
NormanLittle, line 9: Not that it matters _really_ but the question asked for an **8 character** string as the key.

src/selftest/Encode.java
```diff
@@ -0,0 +1,37 @@
...
+		int j;
+
+		System.out.print("Original message: ");
```
NormanLittle, line 12: You could shorten/simplify this output (and the others) to one line with _System.out.printf_ (or _String.format_). 
E.g.

``` java
System.out.printf( "Original message: %s%n", msg );
```

src/selftest/MinMax.java
```diff
@@ -0,0 +1,26 @@
...
+		nums[4] = -978;
+		nums[5] = 5623;
+		nums[6] = 463;
+		nums[7] = -9;
+		nums[8] = 287;
+		nums[9] = 49;
+		
+		min = max = nums[0];
+		for(int num : nums ) {
+			if(num < min) min = num;
```
NormanLittle, line 21: You could use the ternary operator here.  E.g.

``` java
min = num < min ? num : min;
```

self-test.md
alexdeas, line 43: Totally would have worked in Perl.

ardavey, line 43: This is purely an "OMG it's 1am and I need to finish this up" error!  I regularly use the ternary in Perl, and not like this - although it _does_ work like this in Perl.

self-test.md
ardavey, line 57: *checks solutions* ... Apparently not.  God I hate this book.

src/selftest/Encode.java
ardavey, line 9: Seriously, I've got to start reading the whole question!

src/selftest/Encode.java
ardavey, line 12: Noted - this is just a relic of the original from the chapter.

src/selftest/Encode.java
ardavey, line 21: Indeed - more copy-pasta

self-test.md
mocathe1st, line 43: Yeah, it's a shame it doesn't work like that in Java, it'd be cool to be able to put blocks into the ternary to be able to use it instead of if statements. Would make neater code in some cases (and unholy monstrosities in others)

src/selftest/Encode.java
mocathe1st, line 12: Yeah, the printf is pretty cool, I hardly ever use it though because println seems to be burnt into my memory... definitely the way to go though

self-test.md
utopia-2, line 57: Ah, fair enough. Well :+1: for getting it right then.

PR #59, opened 2014-11-27
Chapter 5 self-test answers.

NormanLittle:


src/com/selftest/Encode.java
```diff
@@ -0,0 +1,34 @@
...
+    private static final BiFunction<Integer, String, Integer> cipherCharAt = (a, b) -> {
+        return (b.charAt(a) ^ KEY.charAt(a % KEY.length()));
+    };
+
+    private static String cipher(String message) {
+        StringBuilder result = new StringBuilder();
+        IntStream.range(0, message.length())
+                 .boxed()
+                 .map(i -> (char) cipherCharAt.apply(i, message).intValue())
+                 .forEach(c -> result.append(c));
```
utopia-2: Rather than using `forEach` and an external `StringBuilder`, this could use a joining collector:

``` java
return IntStream.range( 0, message.length() ).boxed()
  .map( i -> (char)cipherCharAt.apply(i, message).intValue() )
  .collect( Collectors.joining() );
```

src/com/selftest/Encode.java
```diff
@@ -0,0 +1,34 @@
...
+
+    public static void main(String[] args) {
+        String message = "This is a test message to encode and decode.";
+        String encoded = cipher(message);
+        String decoded = cipher(encoded);
+
+        System.out.printf("Original message: %s%n", message);
+        System.out.printf("Encoded message : %s%n", encoded);
+        System.out.printf("Decoded message : %s%n", decoded);
+    }
```
utopia-2, line 30: :+1: I like this solution.

src/com/selftest/Encode.java
NormanLittle: Yep, I like that better @Utopia1138 but it doesn't seem to compile. 

```
The method collect(Collector<? super Character,A,R>) in the type Stream<Character> is not applicable for the arguments (Collector<CharSequence,capture#1-of ?,String>)
```

The only way that seems to work is to map the character to a String.  E.g.

``` java
private static String cipher(String message) {
        return IntStream.range(0, message.length()).boxed()
                 .map(i -> Character.toString((char)cipherCharAt.apply(i, message).intValue()))
                 .collect( Collectors.joining() );
    }
```

utopia-2: I wondered about that - didn't actually try my solution though. Unfortunately there doesn't seem to be a collector for composing a `String` from chars. I'd be tempted to remove some of the complexity from the map lambda by re-writing it as:

``` java
private static String cipher(String message) {
        return IntStream.range( 0, message.length() ).boxed()
                 .map( i -> (char) cipherCharAt.apply(i, message).intValue() )
                 .map( Character::toString )
                 .collect( Collectors.joining() );
 }
```

**Edit:** as you point out, because the first map boxes the `char` it causes the `Character::toString` to be ambiguous, so the lambda will have to be explicit about which version to use.

src/com/selftest/Encode.java
```diff
@@ -0,0 +1,34 @@
...
+
+    private static final String KEY = "abcdefgh";
+
+    private static final BiFunction<Integer, String, Integer> cipherCharAt = (a, b) -> {
+        return (b.charAt(a) ^ KEY.charAt(a % KEY.length()));
+    };
+
+    private static String cipher(String message) {
+        StringBuilder result = new StringBuilder();
+        IntStream.range(0, message.length())
```
SimonGross: You could simplify this a bit by using `message.chars()`, which also returns an IntStream. This would simplify `cipherCharAt` slightly as well, as you wouldn't need to call `b.charAt()`

PR #61, opened 2014-11-27
Mcm chapter5

mocathe1st:
Chapter 5 questions and code.


src/mcm/chapter5/AverageFinder.java
```diff
@@ -0,0 +1,21 @@
+package mcm.chapter5;
+
+import java.util.Arrays;
+import java.util.OptionalDouble;
+
+public class AverageFinder {
+
+	public static void main(String[] args) {
+		
+		int[] numbers = {1,2,4,6,7,8,9,4,2,5,6,4,7,3,21,33,20,7,8,10,11};
```
utopia-2, line 10: The question asks this to _...find the average of 10 **double values**_ not ints.

src/mcm/chapter5/Encode.java
```diff
@@ -0,0 +1,33 @@
...
+
+
+public class Encode {
+
+	public String encode( String phrase, String key ) {
+
+		StringBuilder encrypted = new StringBuilder();
+		phrase.chars().forEach( next -> {
+			encrypted.append( (char) (next ^ key.length()) );
+		} );
```
utopia-2, line 11: Nice solution, but I don't think the question intends the key's length to be XOR'ed, or the keys `HiThar` and `secret` will produce equivalent output. I think it wants each `char` of key to XOR the phrase in turn. Single statement lambdas can also be done without braces and a closing `;` if you want. This solution would look like:

``` java
int k = 0;
StringBuilder enc = new StringBuilder();
phrase.chars().forEach( ch -> enc.append( (char)(ch ^ key.charAt(k++ % key.length())) ) );
return enc.toString();
```

src/mcm/chapter5/StringSorter.java
```diff
@@ -0,0 +1,118 @@
...
+		}
+		
+		//Create the result
+		List<String> result = new ArrayList<String>();
+		String next;
+		
+		//Move elements to the right place
+		if(left.get(0).compareTo(right.get(0)) > 0){
+			next = right.get(0);
+			right.remove(0);
```
utopia-2, line 74: The `remove` method just returns the element removed, so this can be simplified:

``` java
next = right.remove(0);
```

src/mcm/chapter5/AverageFinder.java
mocathe1st, line 10: Ah, should have read it properly. Principle is still the same however... :cry:

src/mcm/chapter5/Encode.java
mocathe1st, line 11: Yeah, this is much better. TBH, I rushed this one a bit as I spent far too long getting dragged off on a tangent looking up sorting algorithms for the other question. Quite keen to learn how to use streams and the lambda operator in Java, so trying to squeeze it in where I can :)

src/mcm/chapter5/StringSorter.java
mocathe1st, line 74: Cool, yeah forgot about that. Makes it a bit cleaner

PR #62, opened 2014-11-27
Ch5 gxd

GrahamDavidson:


src/cypher.java
```diff
@@ -0,0 +1,7 @@
+
+
+public class cypher {
+  public static void main (String args[]){
+  	
+  }
```
Spinodal, line 6: so effective it encrypted the source code!

src/average.java
```diff
@@ -0,0 +1,12 @@
+
+
+public class average {
```
SandyThomson, line 3: In Java, the generally accepted way to name your classes (both in terms of filename, and the classname in the file), is to start with a capital letter (so this class and filename would be `Average(.java)` ).

Methods and variables generally start with lower case characters. FYI.

src/minMax.java
```diff
@@ -0,0 +1,15 @@
+
+
+public class minMax {
+  public static void main(String args[]){
+  	int nums[] = {99, -10, 100123, 18, -978, 5623, 463, -9, 287, 49};
+  	int min    = nums[0];
+  	int	max    = nums[0];
+  	
+  	for (int val : nums){
```
SandyThomson, line 9: All good!

src/sortStrings.java
```diff
@@ -0,0 +1,35 @@
...
+
+    // Display original array
+    System.out.print( "Original array is:");
+    for ( String string : data) System.out.print ( " " + string );
+    System.out.println();
+    
+    // This is the Bubble Sort
+    for ( a = 1; a < size ; a++ ) {
+      for ( b = size - 1 ; b >= a ; b -- )  {
+      	if(data[b-1].compareTo( data[b] ) > 0 ){
```
SandyThomson, line 22: All good. It looks a bit odd (IMO) when you start mixing in capital letters, but this works.

src/minMax.java
```diff
@@ -0,0 +1,15 @@
...
+  		if (val < min) min = val;
+  		if (val > max) max = val;
```
mocathe1st, line 11: This is a candidate for the ternary operator, so:
    val = val < min ? min : val;
Although the way you've done it is probably better in this case tbh

src/sortStrings.java
```diff
@@ -0,0 +1,35 @@
+
+
+/*
+ * Sort an array of strings
+ */
+
+public class sortStrings {
```
mocathe1st, line 7: Like Sandy's said elsewhere, general Java convention is to start class names with a capital. The camel casing is good though and is generally used in Java for names (always catches me out in Perl, as I tend to camel case everything :cry: )

PR #65, opened 2014-12-03
spg Chapter 6

SimonGross:
Chapter Six

### DIFF
--- a/src/com/datacash/training/chapter6.md
+++ b/src/com/datacash/training/chapter6.md
@@ -1,0 +1,54 @@
+1) The fragment is not correct, because `count` is private, so it cannot be accessed by a direct reference from outside the class.
+
+2) An access modifier must precede a member's declaration.
+
+3) See QuestionThree.java/Stack.java
+
+4) 
+```java
+void swap( Test one, Test two ) {
+	int placeholder = one.a;
+	one.a = two.a;
+        two.a = placeholder;
+}
+```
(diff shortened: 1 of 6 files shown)


src/com/datacash/training/chapter6.md
```diff
@@ -0,0 +1,54 @@
...
+	one.a = two.a;
+        two.a = placeholder;
+}
+```
+
+5) The fragment is not correct, as overloaded methods cannot be defined by different return types. They must have different arguments.
+
+6)
+```java
+void sdrawkcab( String toReverse ) {
```
codavex, line 20: I read that as ess-draw-kay-cab to begin with. Now I see what you did there.

src/com/datacash/training/chapter6.md
```diff
@@ -0,0 +1,54 @@
...
+
+14) A varargs method can be overloaded.
+
+15)
+```java
+void method( int arg1, int... args ) { //...
+void method( int... args ) { //...
+
+public static void main( String args[] ) {
+	method( 1, 2, 3); // okay
```
codavex, line 51: Is this okay? how do we know if its arg1 followed by two varargs, or three varargs?

SimonGross, line 51: Ah, you're right. When I try to run this code, both calls are marked as ambiguous.

PR #66, opened 2014-12-03
ncl Chapter 6

NormanLittle:
Self Test answers for chapter 6.

### DIFF
--- a/src/com/selftest/Recurse.java
+++ b/src/com/selftest/Recurse.java
@@ -1,0 +1,17 @@
+package com.selftest;
+
+public class Recurse {
+    
+    private static void reverse( String message, int position ) {
+        System.out.print( message.toCharArray()[position] );
+        if (position > 0 ) 
+            reverse( message, --position );
+    }
+    
+    public static void main( String[] args ) {
+        String message = "reverseme";
+        System.out.printf( "%s -> ", message );
+        
+        reverse(message, message.length() - 1);
+    }   
+}
--- a/src/com/selftest/Recurse.java
+++ b/src/com/selftest/Recurse.java
@@ -3,7 +3,7 @@ package com.selftest;
 public class Recurse {
     
     private static void reverse( String message, int position ) {
-        System.out.print( message.toCharArray()[position] );
+        System.out.print( message.charAt( position ) );
         if (position > 0 ) 
             reverse( message, --position );
     }
--- a/src/com/selftest/Stack.java
+++ b/src/com/selftest/Stack.java
@@ -1,0 +1,91 @@
+package com.selftest;
+
+import java.util.Iterator;
+import java.util.NoSuchElementException;
+import java.util.stream.Collectors;
+import java.util.stream.IntStream;
+import java.util.stream.Stream;
+import java.util.stream.StreamSupport;
+
+public class Stack implements Iterable<Character> {
+    private char[] stack;
+    private int capacity;
+    private int size;
+
+    public Stack(int capacity) {
+        this.capacity = capacity;
+        this.stack = new char[this.capacity];
+        this.size = 0;
+    }
+
+    public Stack(Stack other) {
+        this.capacity = other.capacity;
+        this.stack = new char[this.capacity];
+        IntStream.range(0, other.size).forEach(i -> this.push(other.stack[i]));
+    }
+
+    public Stack(char[] values) {
+        this.capacity = values.length;
+        this.stack = new char[this.capacity];
+        for (char c : values)
+            this.push(c);
+    }
+
+    @Override
+    public Iterator<Character> iterator() {
+        return new Iterator<Character>() {
+            @Override
+            public boolean hasNext() {
+                return !isEmpty();
+            }
+
+            @Override
+            public Character next() {
+                return pop();
+            }
+        };
+    }
+
+    public Stream<Character> stream() {
+        return StreamSupport.stream(this.spliterator(), false);
+    }
+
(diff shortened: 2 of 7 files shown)


src/com/selftest/Recurse.java
```diff
@@ -0,0 +1,17 @@
+package com.selftest;
+
+public class Recurse {
+    
+    private static void reverse( String message, int position ) {
+        System.out.print( message.toCharArray()[position] );
```
utopia-2: Not a fan of `String#charAt`?

src/com/selftest/Stack.java
```diff
@@ -0,0 +1,91 @@
...
+import java.util.Iterator;
+import java.util.NoSuchElementException;
+import java.util.stream.Collectors;
+import java.util.stream.IntStream;
+import java.util.stream.Stream;
+import java.util.stream.StreamSupport;
+
+public class Stack implements Iterable<Character> {
+    private char[] stack;
+    private int capacity;
```
utopia-2: This doesn't need to maintain the capacity separately, `stack.length` will have all the information you need.

src/com/selftest/Recurse.java
NormanLittle: doh! That would be easier, yes!

src/com/selftest/Stack.java
```diff
@@ -0,0 +1,91 @@
...
+        };
+    }
+
+    public Stream<Character> stream() {
+        return StreamSupport.stream(this.spliterator(), false);
+    }
+
+    public Stack push(char value) throws IndexOutOfBoundsException {
+        if (size + 1 > capacity)
+            throw new IndexOutOfBoundsException("Error: The stack is full.");
```
utopia-2: The user won't have specified an index, so this might not be the best exception type to use. You could custom some `Stack(Over|Under)flowException` classes, or opt for the existing `Buffer(Over|Under)flowException` types.

src/com/selftest/Stack.java
```diff
@@ -0,0 +1,91 @@
...
+
+        this.stack[size++] = value;
+        return this;
+    }
+
+    public char pop() throws NoSuchElementException {
+        if (isEmpty())
+            throw new NoSuchElementException("Error: The stack is empty.");
```
utopia-2: This exception type is typically used for a search failure rather than an empty container situation. Some kind of underflow exception type would be a good choice.

src/com/selftest/Stack.java
NormanLittle: Fair point.  I did think about using `BufferOverflowException` and `BufferUnderflowException` but wanted to provide an error message.  Thinking about it - no error message is really required as these exceptions are somewhat explicit about what has went wrong.

src/com/trythis/Queue.java
```diff
@@ -0,0 +1,83 @@
...
+import java.util.stream.StreamSupport;
+
+public class Queue<T> implements Iterable<T> {
+    private List<T> queue;
+    
+    public Queue() {
+        this.queue = new ArrayList<T>();
+    }
+    
+    public Queue( Queue<T> other ) {
```
utopia-2: Consider allowing copy construction of a `Queue` of subtypes of `T`, for example:

``` java
public <R extends T> Queue( Queue<R> other ) {
```

src/com/trythis/Queue.java
```diff
@@ -0,0 +1,83 @@
...
+    public Iterator<T> iterator() {
+        return new Iterator<T>() {
+            @Override
+            public boolean hasNext() {
+                return !isEmpty();
+            }
+
+            @Override
+            public T next() {
+                return get();
```
utopia-2: I don't think destructive updates really fit in with the semantics of iterators, a client would not expect:

``` java
Queue<String> q = getAndFillQueue();
for ( String s : q ) {
  System.out.println(q);
}
```

To flush the `Queue` of all contents.

src/com/trythis/QuickSort.java
```diff
@@ -0,0 +1,49 @@
...
+        int i = left;
+        int j = right;
+        
+        T t = items[ (left + right)/2 ];
+        do {
+            while( LESS_THAN( items[i], t ) && i < right ) i++;
+            while( GREATER_THAN( items[j], t ) && j > left ) j--;
+            
+            if ( i <= j )
+                swap(items, i++, j);
```
utopia-2: Won't you need `j--` as well to avoid a redundant comparison on the next loop?

src/com/trythis/Queue.java
NormanLittle: :+1: Nice.  I like that and it could be useful.

src/com/trythis/Queue.java
NormanLittle: That is a very good point which I totally ignored! I think I'll update that because it is just fundamentally wrong.

src/com/trythis/QuickSort.java
NormanLittle: Oops. Typo.

src/com/trythis/Queue.java
```diff
@@ -0,0 +1,74 @@
...
+    public Queue() {
+        this.queue = new ArrayList<T>();
+    }
+    
+    public <R extends T> Queue( Queue<R> other ) {
+        this.queue = new ArrayList<T>();
+        this.queue.addAll( other.queue );
+    }
+
+    public Queue( T[] values ) {
```
utopia-2: You can do the same derived cloning here for convenience:

``` java
public <R extends T> Queue( R[] values ) {
```

PR #67, opened 2014-12-03
Ajd chapter6

alexdeas:
ajd chapter 6

### DIFF
--- a/answers.md
+++ b/answers.md
@@ -1,0 +1,44 @@
+1. The second fragment would not work as it attempts to access a private member in the X class.
+
+2. An access modifier must _precede_ a member's declaration.
+
+3. See Stack.java
+
+4. 
+
+    ```java
+    void swap ( Test a, Test b ) {
+      Test tmp;
+      
+      tmp = a;
+      a = b;
+      b = tmp;
+    }
+    ```
+
+5. The fragment isn't correct as the two methods have the same name
+and arguments. Return type can't be used to distinguish them.
+
+6. See ReverseString.java
+
+7. The variable would be *static*
+
+8. A static block might be used in object initialisation.
+
+9. An inner class is a class defined within another, which will be only available in the enclosing class.
+
+10. *private* is used to make a class member only available within that class.
+
+11. A method's name plus parameter list is known as a *signature*.
+
+12. Call-by-*value*
+
+13. See Args.java
+
+14. Varargs methods can be overloaded, provided that any varargs parameters are last in the parameter list
+and that any preceding arguments are not the same type as the varargs parameter.
+
+15.
+    ```java
+    void method ( int i, int ... varargs ) {}
+    ```
--- a/src/Args.java
+++ b/src/Args.java
@@ -1,0 +1,23 @@
+/*
+ * Demonstrating varargs
+ */
+
+public class Args {
+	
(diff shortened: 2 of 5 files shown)


answers.md
```diff
@@ -0,0 +1,44 @@
...
+12. Call-by-*value*
+
+13. See Args.java
+
+14. Varargs methods can be overloaded, provided that any varargs parameters are last in the parameter list
+and that any preceding arguments are not the same type as the varargs parameter.
+
+15.
+    ```java
+    void method ( int i, int ... varargs ) {}
```
SandyThomson, line 43: This is valid. Where is your second method sig?

src/Args.java
```diff
@@ -0,0 +1,23 @@
...
+    System.out.println( sum( array ) );
+  }
+  
+  static int sum ( int ... array ) {
+    
+    int sum = 0;
+    
+    for ( int i : array ) sum += i;
+  
+    return sum;
```
SandyThomson, line 20: All good :smile: 

I'd _normally_ not put a space between `int` and `...` (so `int...`), but what you have works fine.

src/ReverseString.java
```diff
@@ -0,0 +1,21 @@
...
+    reverseString( "Some string" );
+    System.out.println();
+  }
+  
+  static void reverseString ( String s ) {
+    
+    if ( s.length() > 1 ) {
+      reverseString( s.substring( 1, s.length() ) );
+    }
+    System.out.print( s.substring( 0, 1 ) );
```
SandyThomson, line 18: You did more or less the same as me (except I worked from the other end of the string for some reason). @SimonGross pointed out you could use `.charAt(int)` instead of one of the substring calls, which is a bit more efficient.

src/Stack.java
```diff
@@ -0,0 +1,73 @@
...
+public class Stack {
+  
+  private char[] stack;
+  
+  Stack ( int size ) {
+    this.stack = new char[size];
+  }
+  
+  Stack ( ) {
+    // default to 10 I guess?
```
SandyThomson, line 14: Meh - _force_ the developer to set an initial stack size! :smile:

src/Stack.java
```diff
@@ -0,0 +1,73 @@
...
+    this( 10 );
+  }
+  
+  private int firstIndex () {
+    // Instead of a counter variable, this method figures out the first
+    // empty index (i.e. 0)
+    // I guess someone could insert 0 as a character, but sucks to be them
```
SandyThomson, line 21: Lol - interesting! ...

src/StackDemo.java
```diff
@@ -0,0 +1,39 @@
+/*
+ * Try to use a stack
+ */
+
+
+public class StackDemo {
```
SandyThomson, line 6: In future, instead of doing this, try writing a unit test. It makes it easier to read in future as you don't have to run it to see what it does (see my [StackTest.java](https://github.com/SandyThomson/Java-Training/blob/sdt_chapter6/test/com/selftest/StackTest.java) for an example).

Import JUnit4 into your classpath (it comes with Eclipse), and then you can

alexdeas, line 6: But I wanted a program to run so I could see what it does!

PR #68, opened 2014-12-03
Sandy Thomson Chapter 6 Answers

SandyThomson:


src/com/selftest/Stack.java
```diff
@@ -0,0 +1,33 @@
+package com.selftest;
+
+public class Stack {
+
+	private char[] rawStack;
+	private int maxArrayIndex;
```
SimonGross, line 6: You shouldn't need to store this separately, as you can just call rawStack.length. Although, storing it this way might be easier given the way you're putting things in the array.

src/com/selftest/Stack.java
```diff
@@ -0,0 +1,33 @@
...
+		currentSize = 0;
+		rawStack = new char[initialSize];
+	}
+
+	public void push(char c) {
+		if (currentSize > maxArrayIndex) {
+			throw new RuntimeException("Stack is full, mkay");
+		}
+
+		rawStack[maxArrayIndex - currentSize] = c;
```
SimonGross, line 20: Interesting. If I've got this right you're storing from the end of the array backwards. I think storing things from the beginning of the array is simpler, but this works as well

src/com/selftest/StringBackwards.java
```diff
@@ -0,0 +1,18 @@
+package com.selftest;
+
+public class StringBackwards {
+
+	public static String reverse(String s) {
+
+		if (s.length() > 1) {
+			String stringMinusLastChar = s.substring(0, s.length() - 1);
+			String lastChar = s.substring(s.length() - 1, s.length());
```
SimonGross, line 9: Alternatively, as we're just getting a single character you can use `s.charAt( s.length - 1 )`

src/com/selftest/Stack.java
SandyThomson, line 6: Yeah, I wasn't sure if an array _primitive_ stored the length or not?

src/com/selftest/Stack.java
SandyThomson, line 20: Yeah, don't know why I did it like this. You are probably right, going the other way would probably be more straightforward. For some reason this felt more intuitive to me, but I can't defend it either.

src/com/selftest/StringBackwards.java
SandyThomson, line 9: +1 for the protip :smile:

src/com/selftest/Stack.java
utopia-2, line 6: Arrays of primitives are still objects in Java. They don't provide a whole lot, but you do get a `public final int length` for this purpose.

src/com/selftest/Stack.java
utopia-2, line 20: You could totally argue this models the stack direction of most CPU architectures better ([SO answer](http://stackoverflow.com/a/664779)).

PR #71, opened 2014-12-04
Chapter6 gxd

GrahamDavidson:


src/Answers.md
```diff
@@ -0,0 +1,41 @@
...
+
+Q1) No. Count is a private member variable and cannot be accessed directly from outside of its class.
+
+Q2) Precede
+
+Q3) See [StackDemo.java](./StackDemo.java)
+
+Q4) See [Swap.java](./Swap.java)
+
+Q5) Yes looks fine to me.
```
NormanLittle, line 11: Oops - the fragment is actually incorrect.  This question is to do with **method overloading** and, in this case, only the _return type_ differs between the two method signatures which is insufficient.

src/StackDemo.java
```diff
@@ -0,0 +1,18 @@
...
+	}	
+}	
+
+class Stack {
+  private char data[];
+  private int position=0;
+  
+  Stack (int size) { data = new char[size]; }
+  
+  void push (char ch){ data[position++ % data.length] = ch;   }
```
NormanLittle, line 16: Is this _really_ the behaviour you would expect in the case of exceeding the capacity?  Rolling over is neat but overwriting the bottom of the stack might be a nasty surprise in that case.

src/Swap.java
```diff
@@ -0,0 +1,27 @@
...
+	  System.out.println("alpha = " + alpha.a);
+	  System.out.println("beta = " + beta.a);
+  }
+}
+
+class Test {
+	int a;
+	Test (int i) { a = i; }
+	
+	void swap (Test objA, Test objB){
```
NormanLittle, line 21: It's a little _pedantic_ but a method which swaps the contents of two Test objects (which are both passed as parameters) would make more sense outside of the class (e.g. in the **Swap** class for instance).

src/StackDemo.java
GrahamDavidson, line 16: I have implemented this as a circular stack, if such a thing exists (i.e.like a circular queue). I realise this design is quite primitive however the question was non specific as to the properties required of the stack. As this example was mainly aimed at demonstrating the use of private I thought in the interest of brevity this implementation would suffice.

src/Answers.md
GrahamDavidson, line 11: Ah cool, I get it now, thanks.

src/StackDemo.java
utopia-2, line 16: If this is intended to be a [circular buffer](http://en.wikipedia.org/wiki/Circular_buffer) this is a fairly elegant way of doing it. I suspect it wouldn't fit in with stack semantics though (i.e. growth in one direction).

src/Swap.java
GrahamDavidson, line 21: I was initially tempted to implement it as alpha.swap(beta); but the question sort of suggests that both objects be passed in as parameters. I like your idea best though. I think i was incorrectly tempted in to adding the method to the test class by the way the question was phrased: "Given this class".

PR #74, opened 2014-12-08
Red Chapter 7 exercises

utopia-2:
Try this & exercises for Chapter 7.

### DIFF
--- a/src/org/java/training/exercises/CircleDemo.java
+++ b/src/org/java/training/exercises/CircleDemo.java
@@ -1,0 +1,89 @@
+package org.java.training.exercises;
+
+import java.util.Arrays;
+import java.util.List;
+import java.util.stream.Collectors;
+
+/**
+ * Question 2.
+ */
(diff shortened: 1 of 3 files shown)


src/org/java/training/exercises/CircleDemo.java
```diff
@@ -0,0 +1,89 @@
...
+		super(diameter, "circle");
+	}
+
+	Circle(Circle that) {
+		super(that);
+	}
+
+	@Override
+	public double area() {
+		double r = getWidth() / 2.d;
```
SandyThomson, line 35: As part of our training session, we just happened to observe that the decimal point in `2.d` isn't strictly necessary. You might also get away with omitting `d` too!

We're not angry, we're just disappointed.

:smirk_cat:

utopia-2, line 35: It's not necessary, but it _is_ sexy.

PR #75, opened 2014-12-09
dcox ch07 exercises

codavex:


src/ch07/Circle.java
```diff
@@ -0,0 +1,26 @@
+package ch07;
+
+// A subclass of TwoDShape for circles.
+public class Circle extends TwoDShape {
+
+	static final double PI = 3.14159265359;
```
alexdeas, line 6: Should just round this to 3, much simpler.

src/ch07/self-test.txt
```diff
@@ -0,0 +1,23 @@
+1. No, a superclass does not have access to its subclass' members. Yes, a subclass has access to its superclass' non private members.
+
+2. See Circle.java/AbsShape.java
+
+3. declaring members' private stops subclasses from accessing them
+
+4. super is used to call superclass constructors and to specify a superclass method that has been overridden by its subclass
```
alexdeas, line 7: super is useful for accessing all Superclass members, not just its methods.

src/ch07/self-test.txt
```diff
@@ -0,0 +1,23 @@
...
+
+5. Alpha, then Beta, then Gamma
+
+6. A superclass reference to a subclass object would only have access to the superclass' methods, even if these had been overridden by the subclass.
```
alexdeas, line 11: The question was more about why this is important. While only the superclass methods are available, the subclasses can then handle their own implementation.

src/ch07/self-test.txt
```diff
@@ -0,0 +1,23 @@
...
+
+7. An abstract class is one that hasn't been fully defined, with the implication that subclasses will complete the definition.
```
alexdeas, line 13: Apparently the answer was "a class with one or more abstract methods", which seems a bit self-referential.

src/ch07/self-test.txt
```diff
@@ -0,0 +1,23 @@
...
+
+8. Marking a method final prevents it from being overridden. Marking a class final prevents it from being inherited from.
+
+9. Inheritance, method overriding and abstract classes help support the "one interface, many methods" philosophy aiding polymorphism.
+
+10. The Object class
+
+11. True, a class with an abstract method must be declared abstract. 
+
+12. final is used to make named constants
```
alexdeas, line 23: I like the way `final` answered three of the questions this week.

src/ch07/self-test.txt
utopia-2, line 13: @alexdeas as well as being self-referential, it's also wrong - you're welcome to define a class as `abstract` with no abstract methods. It just stops the class being instantiable.

alexdeas, line 13: No, you're welcome to.

src/ch07/Circle.java
utopia-2, line 6: @Spinodal says use `Math.PI`.

PR #76, opened 2014-12-09
ncl Chapter 7

NormanLittle:


answers.md
```diff
@@ -0,0 +1,33 @@
...
+* access a member from the superclass from a subclass
+```java
+super.name;
+```
+
+5) _Alpha_, _Beta_ then _Gamma_.
+
+6) When a overridden method is called on a superclass reference the subclass version will be executed.
+
+7) An **abstract** class is a class which contains at least _one_ abstract method.  It is often used to define common fields and/or behaviour and cannot be instantiated directly.
```
utopia-2, line 23: It can have zero abstract methods if it likes. There's not a lot of use for it, but you may want to make the class un-instantiable.

answers.md
```diff
@@ -0,0 +1,33 @@
...
+
+5) _Alpha_, _Beta_ then _Gamma_.
+
+6) When a overridden method is called on a superclass reference the subclass version will be executed.
+
+7) An **abstract** class is a class which contains at least _one_ abstract method.  It is often used to define common fields and/or behaviour and cannot be instantiated directly. 
+
+8) Declare both as **final**.
+
+9) _Inheritance_, _method overriding_ and _abstract classes_ are used to support _polymorphism_ by allowing the user to create a common interface which can be implemented by many subclasses.  The reference to just the superclass can, therefore, refer to many possible implementations.
```
mocathe1st, line 27: :+1: Nice answer

PR #77, opened 2014-12-09
mssw Chapter 7

Spinodal:


src/main/java/Circle.java
```diff
@@ -0,0 +1,45 @@
...
+	 * @param n
+	 */
+	public Circle(double x, String n) {
+		super(x, n);
+	}
+
+	/**
+	 * @param ob
+	 */
+	public Circle(TwoDShape ob) {
```
utopia-2, line 30: If you pass in a `TwoDShape` type without equal width/height, your area calculation below will be incorrect.

PR #78, opened 2014-12-10
MCM Chapter 7

mocathe1st:


src/chapter7_answers.markdown
```diff
@@ -0,0 +1,27 @@
...
+
+**Q1**: A superclass does not have access to the members of a subclass (unless they are overriden versions of the superclass's members). A sub class does have access to members of its superclass given providing they are not defined with the private access modifier.
+
+**Q2**: See Ellipse.java
+
+**Q3**: Using the private access modifier restricts access to members of a superclass
+
+**Q4**: The purpose of the two version of super are:
+* To initiate the constructor of the super class, by using `super()` in the subclass's constructor
+* To access members of the super class, by using `super().membername`
```
NormanLittle, line 11: I expect it is a typo but that won't compile as `super()` calls the constructor of the superclass which is only allowed in the subclass constructor.  Should just use the **super** keyword instead. E.g. `super.membername`.

src/chapter7_answers.markdown
```diff
@@ -0,0 +1,27 @@
...
+
+**Q5:**: Constructors will complete execution in the following order: Alpha, Beta, Gamma
+
+**Q6**: This is important as subclasses can be substituted in a reference to their superclass, this can modify the behaviour of an application at runtime when methods have been overridden.
+
+**Q7**: An abstract class is a specification class which contains some functionality but cannot be instantiated as of itself. Instead it should be extended by an implementing class and required methods overriden. It can be thought of as like an interface that includes some functionality, although there are key differences as sub classes extend rather than implement abstract classes.
```
NormanLittle, line 17: :+1: although interfaces do now allow functionality with _default_ methods.

src/mcm/chapter7/Ellipse.java
```diff
@@ -0,0 +1,64 @@
...
+		double area = Math.PI * axisARadius * axisBRadiues;
+		
+		return area;
+	}
+	
+	/**
+	 * Print out dimensions - show width, height and area
+	 */
+	@Override
+	public void showDim() {
```
NormanLittle, line 43: I'm being a bit _pedantic_ here but you are technically giving the `showDim()` method an additional behaviour in that it now displays area too.  I would prefer to see that remain consistent for all subclasses.

src/mcm/chapter7/Triangle.java
```diff
@@ -0,0 +1,19 @@
...
+
+// A subclass of TwoDShape for triangles.
+class Triangle extends TwoDShape {
+
+	private String style;
+
+	// A default constructor.
+	Triangle() {
+		super();
+		style = "none";
```
NormanLittle, line 11: What is `style` used for?

src/chapter7_answers.markdown
mocathe1st, line 11: Yeah, that's correct, the brackets on the keyword are wrong... to much hurry or something :s

src/chapter7_answers.markdown
mocathe1st, line 17: Oh, didn't know that.... Is that as of Java 8?

src/mcm/chapter7/Triangle.java
mocathe1st, line 11: No idea, TBH I just copied this one from the book and then realised there was no real point to it... didn't actually intend to push it to git. I think style was supposed to show that you can add new behaviour to a subclass, but not really sure as it isn't a great example.

src/chapter7_answers.markdown
NormanLittle, line 17: Yeah - a new feature in Java 8.

src/mcm/chapter7/Triangle.java
NormanLittle, line 11: Ah, sorry - forgot it was somewhere in the chapter.  I ended up removing all the stuff I didn't think was necessary.

PR #80, opened 2014-12-10
Chapter 7 answers

SandyThomson:


fitzretsil:
Looks good to me, I especially like the Pentagon

src/com/selftest/Circle.java
```diff
@@ -0,0 +1,19 @@
+package com.selftest;
+
+import java.lang.Math;
+
+public class Circle extends TwoDShape {
+
+	private int radius;
```
utopia-2, line 7: We were perusing your code and wondered why you included this as another member when the information is contained in the `super.width / 2`.

PR #84, opened 2014-12-16
Chapter 8 Self Test Answers

NormanLittle:
ncl Chapter 8

### DIFF
--- a/answers.md
+++ b/answers.md
@@ -35,7 +35,7 @@ Queue<String> fixedQueue = new FixedQueue<String>( 10 );
  
  8) Yes using the **extends** keyword.
  
- 9) See [IVehicle.java](src/com/selftest/Vehicle/IVehicle.java)
+ 9) See [IVehicle.java](src/com/selftest/vehicle/IVehicle.java)
  
  10) Yes because they are also implicitly **public**.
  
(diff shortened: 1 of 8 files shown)


answers.md
```diff
@@ -0,0 +1,53 @@
...
+Queue<String> fixedQueue = new FixedQueue<String>( 10 );
+```
+
+ 6) The use of an _interface_ to define the methods which implementing subclasses **need** to provide (or _optionally_ can provide in the case of **default** methods).
+ 
+ 7) There is no limit how many classes can implement a particular _interface_ or how many _interfaces_ a class can implement.
+ 
+ 8) Yes using the **extends** keyword.
+ 
+ 9) See [IVehicle.java](src/com/selftest/Vehicle/IVehicle.java)
```
Spinodal: Link's broken, capital V in Vehicle

NormanLittle: Thanks.  Updated.

answers.md
```diff
@@ -0,0 +1,53 @@
...
+ 
+ 10) Yes because they are also implicitly **public**.
+ 
+ 11) True.
+ 
+ 12) **java.lang**.
+ 
+ 13) **default**.
+ 
+ 14) See [Queue.java](src/com/selftest/qpack/Queue.java)
```
Spinodal: I dont actualy see a static method in there, so i'm guessing you accidentally skipped question 14 and numbered 15 and 16 wrong :stuck_out_tongue:

NormanLittle: Oops.  Less hurry, more speed :)

PR #85, opened 2014-12-16
ajd answers for chapter 8

alexdeas:
ajd answers for chapter 8

### DIFF
--- a/answers.md
+++ b/answers.md
@@ -1,0 +1,33 @@
+1. See IQueueDemo.java (imagine the other classes are there too)
+
+2. A namespace is a collection of classes or interfaces where any class or method names must be unique.
+
+3. Packages are stored in directories.
+
+4. =protected= prevents access to a class member outside the class and inheritance hierachy. The default
+access allows access to a class member within its package.
+
+5. Members of a package can be used by other packages by fully qualifying their namespace, or by
+importing the package.
+
+6. Interfaces?
+
+7. As many as you like.
+
+8. Interfaces can be extended by other interfaces. Classes may not extend interfaces.
+
+9. See IVehicle.java
+
+10. They can be shared by other parts of a program as they are also implicitly public.
+
+11. True.
+
+12. java.lang
+
+13. =default=
+
+14. They did add static methods to interfaces in Java 8. Here's to functional programming.
+
+15. See ICharQ.java.
+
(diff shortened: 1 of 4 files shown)


src/IQueueDemo.java
```diff
@@ -0,0 +1,81 @@
+
+
+public class IQueueDemo {
+  public static void main(String args[]) {    
+    
+    // The extent of question 1
+    qpack.FixedQueue q1 = new qpack.FixedQueue(10);
```
SandyThomson, line 7: Nice. Also note you _could_ have also used import statements, which would mean you don't always have to qualify what package the class is in.

src/iVehicle.java
```diff
@@ -0,0 +1,18 @@
+
+
+
+public interface iVehicle {
+   
+  int range();
+  double fuelneeded(int miles);
```
SandyThomson, line 7: I actually made these two methods default methods. Didn't _need_ to though.

answers.md
```diff
@@ -0,0 +1,33 @@
+1. See IQueueDemo.java (imagine the other classes are there too)
+
+2. A namespace is a collection of classes or interfaces where any class or method names must be unique.
+
+3. Packages are stored in directories.
+
+4. =protected= prevents access to a class member outside the class and inheritance hierachy. The default
+access allows access to a class member within its package.
```
SandyThomson, line 8: I've read this answer a few times. I'm reasonably sure you know what's going on, but I'll restate my understanding of this.

`protected` is less restrictive than `default`, as it allows classes outside the package to access members iff they are subclasses of the class with the protected members..

answers.md
```diff
@@ -0,0 +1,33 @@
...
+
+5. Members of a package can be used by other packages by fully qualifying their namespace, or by
+importing the package.
+
+6. Interfaces?
+
+7. As many as you like.
```
SandyThomson, line 15: Sort of true, but see my answer for utter pedantry.

answers.md
```diff
@@ -0,0 +1,33 @@
...
+
+10. They can be shared by other parts of a program as they are also implicitly public.
+
+11. True.
+
+12. java.lang
+
+13. =default=
+
+14. They did add static methods to interfaces in Java 8. Here's to functional programming.
```
SandyThomson, line 29: Also interestingly, you can't override the static method, if I recall correctly..

PR #96, opened 2015-01-09
ready for review

GrahamDavidson:


src/StackDemo.java
```diff
@@ -0,0 +1,41 @@
...
+		for(char ch : data) stack.push(ch);
+		try{
+		  for (int i = 0; i < 11; i++) System.out.println("Character " + i + " = "+stack.pop());
+		}
+		catch (StackUnderflowException e){
+		  System.out.println("Stack Underflow value no character could be returned.");
+		}
+	}	
+}	
+class StackOverflowException extends Throwable {}
```
NormanLittle, line 15: Would have been neater to construct your exceptions with the specific error message rather than relying on your catch to display the details of what went wrong.

src/StackDemo.java
```diff
@@ -0,0 +1,41 @@
...
+class Stack {
+	
+  private char data[];
+  private int position=0;
+  
+
+  
+  Stack (int size) { data = new char[size]; }
+  
+  void push (char ch)
```
NormanLittle, line 27: Would have preferred to see this method throw the exception rather than throw it, catch it and display an error message.

src/answers.md
```diff
@@ -0,0 +1,18 @@
...
+
+ 1) Throwable.
+ 2) The try block should surround the code for which exceptions are to be handled.
+    Immediately following the try block the catch block(s) provide code to handle various types of exception which may be thrown within the scope of the try block.
+ 3) There is no try block preceeding the catch block.
+ 4) The exception will be handled by the JVM default behaviour is to terminate program execution and display a stack trace.  
+ 5) Since B is a subclass of A the catch block for B will be unreachable code and will cause a compilation error.
+ 6) Yes
+ 7) False in some cases finally may be the last block executed but this is not always the case. If for example an exception is subsequently caught by an enclosing catch the finally block would not be the last code to be executed.
+ 8) Checked exceptions must be explicitly declared in a throws clause.
```
NormanLittle, line 18: Might have been worth expanding this to define what makes an exception **checked**.

PR #99, opened 2015-01-13
dcox ch09

codavex:


src/ch09/self-test.txt
```diff
@@ -0,0 +1,30 @@
...
+
+4. Eventually, the exception will be passed to the JVM which will terminate.
+
+5. B will not be caught, as the superclass A block will match all subclasses.
+
+6. Yes, an exception can be thrown to an outer catch.
+
+7. False, finally blocks are only executed after a try/catch block, it doesn't necessarily terminate execution.
+
+8. All except RunTimeError exceptions.
```
alexdeas, line 15: `Error` exceptions too.

src/ch09/self-test.txt
```diff
@@ -0,0 +1,30 @@
...
+10. See code
+
+11. Exceptions can be thrown by
+    - the JVM,
+    - errors in code, or
+    - generated and thrown by code.
+
+12. Error and Exception or the direct subclasses of Throwable
+
+13. Multi catch allows you to specify an exception that caused an exception.
```
alexdeas, line 28: Multi catch is where you can specify more than one exception type for one catch block.

src/ch09/Stack.java
```diff
@@ -0,0 +1,139 @@
+
+package ch09;
+
+class StackFullException extends Exception {
+	// not sure if the default value is better than suppressing the warning about it not being declared?
+	private static final long serialVersionUID = 1L;
+
```
alexdeas, line 7: Might have been handy to have some sort of stringified version of the exception.

PR #102, opened 2015-01-14
Andy chapter 9

fitzretsil:


answers/chapter9.md
```diff
@@ -0,0 +1,94 @@
...
+// ...
+
+try {
+  // ...
+ }
+catch (A exc) { ... }
+catch (B exc) { ... }
+``` 
+
+The second catch will never be used as exceptions of type B are also of type A, due to B being a subclass of A.
```
alexdeas, line 52: Also unreachable code panic!

answers/chapter9.md
```diff
@@ -0,0 +1,94 @@
...
+
+A problem with the JVM, a problem with the code written or explicitly using the **throw** statement.
+
+12) _What are two direct subclasses of **Throwable**?_
+
+**Exception** and **Error**
+
+13) _What is the multi-catch feature?_
+
+Alongs a single catch statement to handle multiple types of exceptions
```
alexdeas, line 90: Alongs?

src/Stack.java
```diff
@@ -0,0 +1,50 @@
...
+		tos = ob.tos;
+		stck = new char[ob.stck.length];
+
+		// copy elements
+		for (int i = 0; i < tos; i++)
+			stck[i] = ob.stck[i];
+	}
+
+	// Construct a stack with initial values.
+	Stack(char a[]) throws StackFullException {
```
alexdeas, line 23: Do you need to specify that the constructor throws the exception, since it's thrown by the push() method?

src/StackEmptyException.java
```diff
@@ -0,0 +1,12 @@
+public class StackEmptyException extends Throwable {
```
alexdeas, line 1: You subclassed Throwable here rather than Exception. Seems like it could mean more work in real world examples.

src/StackEmptyException.java
```diff
@@ -0,0 +1,12 @@
...
+
+	/**
+	 * 
+	 */
+	private static final long serialVersionUID = 8940747433345773938L;
+
+	public StackEmptyException() {
+		System.out.println("Stack is empty");
```
alexdeas, line 9: Would you normally want an exception to include functionality? I thought they were mostly concerned with detailing information about whatever caused the exception and handling them was left to catch blocks.

PR #104, opened 2015-01-14
mssw chapter 9

Spinodal:


answers.md
```diff
@@ -0,0 +1,102 @@
...
+
+```java
+// ...
+vals[18] = 10;
+catch (ArrayIndexOutOfBoundsException exc) {
+	// handle error
+} 
+```
+
+The comment is stating that it is handling an error, when in fact it is handling an exception.
```
utopia-2, line 23: It states _handle error_ with a lowercase error, i.e. not the `Error` class. It's valid to refer to catching exceptions as a means of error handling.

answers.md
```diff
@@ -0,0 +1,102 @@
...
+
+## 9. What is wrong with this fragment?
+
+```java
+class MyClass { // ... }
+// ...
+throw new MyClass();
+```
+
+A line comment was used in the first `// ...` meaning that the class `MyClass` does not have a closing bracket.
```
utopia-2, line 78: :+1: the book's overuse of `// ...` to indicate _other stuff here_ is kind of annoying.

src/main/java/MikeException.java
```diff
@@ -0,0 +1,57 @@
+/**
+ * 
+ */
+package main.java;
+
+/**
+ * @author Mike
+ *
+ */
+public class MikeException extends Exception {
```
utopia-2, line 10: The question asked for _exceptions_, i.e. a type for each case. `MikeException` doesn't really indicate what went wrong to the caller (unless you're manually computing the results behind the scenes).

PR #105, opened 2015-01-15
mcm Chapter 9

mocathe1st:


src/mcm_chapter_9/mcm_ch9_answers.markdown
```diff
@@ -0,0 +1,50 @@
...
+	catch(Exception e){
+		//Catch all exceptions
+	}
+````
+
+**Q3**: It doesn't have a try block, so won't catch the error. In fact I'm not sure it'll even compile with no try block, but even if it does - its bad news.
+
+**Q4**: If an exception is not caught then a runtime error will occur and the Java program will stop. Exception handling allows for graceful failure to avoid the program crashing.
+
+**Q5**: Exception "B" will never be caught as it is a subclass of "A" and since "A" is caught first any instance of "B" will be caught by the catch for "B". The fix for this is to change the ordering, so that "B" is caught first.
```
NormanLittle, line 22: I'm a little _confused_.  'B' exceptions will be caught - they will just be caught by the 'A' catch block because it is defined before the 'B' catch block. As things stand - 'B' exceptions will **not** be caught by the catch for 'B'.  

The fix is correct though - the ordering needs to be switched :+1:

src/mcm_chapter_9/mcm_ch9_answers.markdown
```diff
@@ -0,0 +1,50 @@
...
+
+**Q6**: Yes, an inner catch can thrown an exception to an outer catch.
+
+**Q7**: Not true. The finally block is **ALWAYS** run after a try / catch block regardless of whether an exeption occurred. Commonly used to clean up open resources, etc. The rest of the program will continue to function after the finally block.
+
+**Q8**: Checked exceptions need to be explicitly declared in a throws clause, as the compiler checks if they are thrown. Unchecked exceptions do not need to be declared in a throws clause.
```
NormanLittle, line 28: It might have been worth specifying what makes an exception **checked**.

src/mcm_chapter_9/mcm_ch9_answers.markdown
```diff
@@ -0,0 +1,50 @@
...
+
+**Q9**: MyClass does not extend Throwable so cannot be thrown.
+
+**Q10**: See Java code
```
NormanLittle, line 32: What, no link? ;)

src/mcm_chapter_9/mcm_ch9_answers.markdown
```diff
@@ -0,0 +1,50 @@
...
+
+**Q12**: Exception and Error
+
+**Q13**: The multi catch feature was introduced in Java 7 to allow a catch block to catch more than one exception using the following syntax:
+
+```java
+	try{
+		//..do something
+	}
+	catch(ExceptionA a | ExceptionB b){
```
NormanLittle, line 44: Oops - don't need a variable for each exception.  E.g.

``` java
catch( ExceptionA | ExceptionB e ) {
 ...
```

src/mcm_chapter_9/stack/Stack.java
```diff
@@ -0,0 +1,60 @@
...
+	}
+	
+	//Construct a stack with initial values.
+	public Stack(char a[]) {
+		stck = new char[a.length];
+		for(int i = 0; i < a.length; i++) {
+			try {
+				push( a[i] );
+			}
+			catch ( StackFullException e ) {
```
NormanLittle, line 35: Would probably be better to just throw the exception, no?

src/mcm_chapter_9/mcm_ch9_answers.markdown
mocathe1st, line 22: True, this is a fairly badly worded answer.. I think I knew what I meant in my head, but didn't express it very well. Yeah, B exceptions will be caught, just not by the block specifically looking for B exception.s

PR #106, opened 2015-01-16
Added ch9 answers

ardavey:
For review


selftest.txt
```diff
@@ -0,0 +1,7 @@
...

A: Error and Exception

13. What is the multi-catch feature?

A: Allows a single catch{} block to trap multiple exceptions.

14. Should your code typically catch exceptions of type Error?
+
+A: No; These relate to JVM errors which are beyond our control.
```
Spinodal, line 7: "...beyond our control."

unless someone were to create a custom subclass of error, but that would be strange....

src/selftest/StackDemo.java
```diff
@@ -0,0 +1,98 @@
...
+	}
+}
+
+class StackEmptyException extends Exception {
+	public String toString() {
+		return "\nStack is empty";
+	}
+}
+
+class Stack {
```
Spinodal, line 21: there are no access modifiers on your methods!!! raaage!!

PR #112, opened 2015-01-22
ready for review

GrahamDavidson:


Answers.md
```diff
@@ -0,0 +1,57 @@
...
+
+	FileReader in = new FileReader("inFile.txt");
+
+5)	
+
+	RandomAcessFile in = new RandomAcessFile("inFile.bin", "rw");
+
+6) 
+
+	double d = Double.parse("123.23");
```
utopia-2, line 25: For whatever reason, they opted for `T.parseT(String)` for the primitive wrapper types, though `T.parse` makes more sense to me. This would need to be `Double.parseDouble`.

utopia-2:
:+1: Looks good.

PR #113, opened 2015-01-22
mcm chapter10

mocathe1st:


src/mcm_chapter_10/mcm_ch10_answers.markdown
```diff
@@ -0,0 +1,70 @@
...
+```java
+ 	RandomAccessFile raf = new RandomAccessFile("random.dat", "rw")
+ 
+ 	raf.seek(0);
+	d = raf.readDouble(); //etc...
+```
+
+**6. How can you convert a numeric string such as "123.23" into its binary equivalent?**
+
+There are a number of ways of converting a string number into binary (e.g., see [ConvertStringToBinary.java](io/ConvertStringToBinary.java), however easier ways are to use the Integer Parsing methods e.g.,
```
utopia-2, line 34: `ConvertStringToBinary.java` appears to convert a string into a binary _string_, rather than into some primitive data format. An example of doing this by hand (inefficiently in this case) might be:

``` java
    int parseInt(String num) {
      int out = 0;
      for (int i = num.length()-1; i >= 0; --i) {
        char dec = num.charAt(i);
        if ( i == 0 && dec == '-' ) {
          out = -out;
          break;
        }
        if ( dec < '0' || dec > '9' ) throw new NumberFormatException("Non-integer character found");
        out += (dec - '0') * (Math.pow(10, num.length() - 1 - i));
      }

      return out;
    }
```

PR #116, opened 2015-01-27
ncl Chapter 11

NormanLittle:


answers.md
```diff
@@ -0,0 +1,31 @@
...
+  MyThrd.setPriority( Thread.NORM_PRIORITY + 3 );
+```
+
+6)  The effect is that the method will only be able to be executed by one thread at a time for a given object of that class type.  When the method is called the calling thread will enter the object's monitor which locks the method which prevents any other thread entering the method or any other _synchronized_ method defined by the class.
+
+7)  Thread communication.  
+
+8)  See [Clock.java](src/com/selftest/Clock.java)
+
+9)  These methods are _deprecated_.
```
utopia-2, line 25: _Why_ are they deprecated?

src/com/selftest/queue/AbstractQueue.java
```diff
@@ -0,0 +1,56 @@
...
+
+	@Override
+	public synchronized Iterator<T> iterator() {
+		return new Iterator<T>() {
+
+			int position = 0;
+
+			@Override
+			public boolean hasNext() {
+				return position < queue.length;
```
utopia-2: Won't this iterate over a load of empty (or assumed deleted) indices by not considering `putloc` and `getloc`? Based on what this iterator does, you could have just returned: `return Arrays.asList(queue).iterator()`.

src/com/selftest/clock/Clock.java
```diff
@@ -0,0 +1,54 @@
...
+		state = s;
+		if ( !running ) {
+			notify();
+		}
+		else {
+			state.run();
+
+			notify();
+			try {
+				while ( state != (s == State.TICKED ? State.TOCKED : State.TICKED) )
```
utopia-2: Why not just `while( state == s )`?

src/com/selftest/queue/AbstractQueue.java
NormanLittle: Yes, it will as my intention was to show the empty/deleted indices too (for no other reason than it being strangely satisfying to see the queue half full/empty).  

Fair point - I could/should have just used the List::iterator() method.

src/com/selftest/clock/Clock.java
NormanLittle: Doh! Yes, that would have been far cleaner.

src/com/selftest/queue/AbstractQueue.java
```diff
@@ -0,0 +1,56 @@
...
+			}
+
+			@Override
+			public T next() {
+				return queue[position++];
+			}
```
utopia-2: Are `T` writes guaranteed to be atomic? What will happen if this reads from the queue whilst another thread writes to it at this location?

utopia-2: As it happens, Java mandates reference writes be atomic [[1]](http://docs.oracle.com/javase/specs/jls/se5.0/html/memory.html#17.7), so this is a safe iterator :smile:.

answers.md
NormanLittle, line 25: _Why_ _Not_? :-)

Okay, the answer probably required here is that **suspend** and **stop** are both deprecated because they can cause problems (deadlock etc) and **resume** is deprecated because it cannot be used without **suspend**.

utopia-2, line 25: Correct; your prize: :moneybag:.

src/com/selftest/queue/AbstractQueue.java
NormanLittle: Yep, that is _totally_ what I thought.  Honestly. Maybe.  Well, okay - no, I hadn't thought about it at all and just got lucky.

answers.md
```diff
@@ -0,0 +1,31 @@
...
+
+10)  **getName()**.
+
+11)  A **boolean**.  Specifically, it will return **true** if the thread is alive (e.g. thread is still running) and **false** if not (e.g. thread has terminated).  
+
+12)  See [QueueDemo.java](src/com/selftest/QueueDemo.java)
```
Spinodal, line 31: I don't see anything particularly thread safe in this class, did you link to the right one? or is the thread safe-ness hidden behind other things?

NormanLittle, line 31: I linked to the demo class for brevity assuming people might look at the classes being used (E.g. Queue, AbstractQueue, FixedQueue etc) rather than listing them all specifically.

PR #118, opened 2015-01-28
Red Chapter 11

utopia-2:


answers.md
```diff
@@ -0,0 +1,76 @@
...
+
+ 11) _What does `isAlive()` return?_
+
+A boolean determining whether the thread is active (that is, started and not complete,
+the thread doesn't need to be in context to be active).
+
+ 12) _On your own, try adding synchronization to the `Queue` class developed in
+      previous chapters so that it is safe for multithreaded use._
+
+On my own? Who had I been answering the other questions with!?
```
NormanLittle, line 74: Easy, it's not worth it...just walk away!

src/org/java/training/SynchQueueDemo.java
```diff
@@ -0,0 +1,136 @@
...
+		head.next = head.next.next;
+		return item;
+	}
+	
+	public synchronized Iterator<T> iterator() {
+		return new Iterator<T>() {
+			Node index = head.next.next;
+
+			public boolean hasNext() {
+				synchronized(LinkedQueue.this) {
```
NormanLittle, line 123: So, given reference writes will be atomic, does this mean **synchronized** is not required here?

utopia-2, line 123: This still needs to be `synchronized` - dereference of `head`, then checking address of `head.next`. Between the dereference and the field access, either could be overwritten.

PR #130, opened 2015-02-04
Dcox chapter12

codavex:


src/TrafficLightSimulator.java
```diff
@@ -0,0 +1,104 @@
...
+				System.out.println( exc );
+			}
+			changeColor();
+		}
+	}
+
+	// Change color. 
+	synchronized void changeColor() {
+		// get the values, then add one, rolling over if necessary
+		TrafficLightColor[] values = TrafficLightColor.values();
```
alexdeas, line 59: Ah, I didn't manage to figure out how to retrieve an enum based on the ordinal value. Hadn't thought about using an array.

src/EnumOrdinals.java
```diff
@@ -0,0 +1,17 @@
...
+} 
+
+public class EnumOrdinals {
+ public static void main(String args[])  
+ { 
+   // Obtain all ordinal values using ordinal(). 
+		System.out.println( "Here are all Tools constants" +
+                      " and their ordinal values: "); 
+		for ( Tools t : Tools.values() )
+     System.out.println(t + " " + t.ordinal());
```
SandyThomson, line 13: Avoided explicit `.toString()`. Winner!

PR #133, opened 2015-02-04
Chapter 12 answers

SandyThomson:


answers/chapter12.md
```diff
@@ -0,0 +1,43 @@
+  1. The constants in an Enumeration have the type of the enumeration itself.
+  
+  2. java.lang.Enum?
```
codavex, line 3: Are you asking me or telling me? :stuck_out_tongue_winking_eye:

answers/chapter12.md
```diff
@@ -0,0 +1,43 @@
...
+ making your code potentially more readable, especially when used repeatedly.
+ 
+ 9. You generally want to use static import to cut down on clutter. If it's a one off use you probably don't need to bother.
+ You can run into collisions, however from memory the compiler will alert you to this fact and force you to fully qualify the use.
+ 
+ 10. An annotation is syntactically based on a/an interface
+ 
+ 11. A marker annotation is one that contains no values (e.g @Expose, @Deprecated, @Override etc)
+ 
+ 12. An annotation can be applied to Classes, Methods and Class variables. Java 8 might have extended this, although I didn't really understand what the book meant here.
```
codavex, line 43: Looks good :+1:

src/com/selftest/trafficlight/TrafficLightColour.java
```diff
@@ -0,0 +1,17 @@
+package com.selftest.trafficlight;
+
+public enum TrafficLightColour {
+	RED(1), GREEN(1), AMBER(1);
```
codavex, line 4: Wow, you don't get much time on that green light, do you?

Good move to the more accurate AMBER instead of YELLOW :+1:

src/com/selftest/trafficlight/TrafficLightSimulator.java
```diff
@@ -0,0 +1,62 @@
...
+				System.out.println(e);
+			}
+			changeColour();
+		}
+		
+	}
+
+	public synchronized void changeColour() {
+		// Excuses to use ordinals etc
+		tlc = (tlc.ordinal() == 2) ? TrafficLightColour.values()[0]
```
codavex, line 36: I like this, I used a modulus operator, but it looked clunky (to me)

test/com/selftest/ToolsTest.java
```diff
@@ -0,0 +1,14 @@
+package com.selftest;
+
+import org.junit.Test;
+
+public class ToolsTest {
+
+	@Test
+	public void TestTools() {
+		for (Tools t : Tools.values()) {
+			System.out.println("Enum value:" + t.toString());
```
codavex, line 10: Think you can get away with just t here, don't need the toString

src/com/selftest/trafficlight/TrafficLightSimulator.java
WayneBooth, line 36: But what happens when Oracle realises their's more to the world than just U.S.A.? And are forced kicking and screaming to add the colour "RED_n_AMBER"? 
You'll have to change this code.

codavex, line 36: True, my modulus looked clunky in part due to the call to length, which would be needed here to future proof it.

The whole thing would need updated to refer to the on-off state of each bulb in the light box

(RED_n_AMBER's a sort of murky orange, isn't it?)

PR #136, opened 2015-02-05
mssw chapter 12

Spinodal:


answers.md
```diff
@@ -0,0 +1,80 @@
+# mssw Chapter 12 Answers
+
+## 1. Enumeration constants are said to be *self-typed*. What does this mean?
+
+I didn't really understand this but i think they mean that the type of the enum is the class name that it is contained in.
```
utopia-2, line 5: Yeah, it more or less just means _what type do the enum constants have, given that they're not explicitly typed_.

answers.md
```diff
@@ -0,0 +1,80 @@
...
+
+Autoboxing will automatically convert a primitive into its associated object wrapper class without the programmer needing to do anything special.
+
+## 6. Change the following fragment so that is uses autoboxing?
+
+```java
+Short val = new Short(123);
+```
+
+really???
```
utopia-2, line 38: Yup. :octopus:

answers.md
```diff
@@ -0,0 +1,80 @@
...
+So when should you use static import? Very sparingly! Only use it when you'd otherwise be tempted to declare local copies of constants, or to abuse inheritance (the Constant Interface Antipattern). In other words, use it when you require frequent access to static members from one or two classes. If you overuse the static import feature, it can make your program unreadable and unmaintainable, polluting its namespace with all the static members you import. Readers of your code (including you, a few months after you wrote it) will not know which class a static member comes from. Importing all of the static members from a class can be particularly harmful to readability; if you need only one or two members, import them individually. Used appropriately, static import can make your program more readable, by removing the boilerplate of repetition of class names.
+
+Copyright &copy; 2004, 2010 Oracle and/or its affiliates. All rights reserved.
```
utopia-2, line 64: ...We've lost @Spinodal :finnadie:. Another victim of Java Training.

answers.md
```diff
@@ -0,0 +1,80 @@
...
+
+It imports the static java.lang.Integer.parseInt; method.
+
+## 9. Is static import designed for special case situations, or is if good practice to bring all static members of all classes into view?
+
+its not good practice to import all static members into your class as it will pollute the namespace. Ideally you would only import static members that you use very frequently throughout your class. It would also be best to import each method individually so that it is clear where the method is coming from.
+
+## 10. An annotation is syntactically based on a/an \_\_\_\_\_\_\_\_\_\_\_
+
+For this and the next few answers, please [buy my book](xtra-answers.md): *Java: The Complete Answers, Twelfth Edition* (Prophet Newspaper/Wambeek Home Edition, 2015)
```
utopia-2, line 80: Nice of you to link to a free edition. :+1:

src/main/java/traffic/TrafficLightColour.java
```diff
@@ -0,0 +1,15 @@
+package main.java.traffic;
+
+//An enumeration of the colors of a traffic light. 
+enum TrafficLightColour {
+	RED(1000), RED_YELLOW(500), GREEN(1000), YELLOW(1000);
```
utopia-2, line 5: Seems like an awkward way of getting around the book version's lack of amber after green. Given that these are public fields, it could be confusing to somone using it. You could get around it using a private static array of state transitions and a method providing an iterator over them?

``` java
private static final TrafficLightColour[] STATE_TRANSITIONS = { RED, YELLOW, GREEN, YELLOW };

public static Iterator<TrafficLightColour> iterator() {
  return new Iterator<TrafficLightColour>() {
    int i = 0;
    boolean hasNext() { return true; }
    TrafficLightColour next() { return STATE_TRANSITIONS[i = ++i % STATE_TRANSITIONS.length]; }
  };
}
```

I haven't tested this though, it might complain about using constants before they've been loaded.

src/main/java/traffic/TrafficLightSimulator.java
```diff
@@ -0,0 +1,78 @@
...
+		thrd = new Thread(this);
+		thrd.start();
+	}
+
+	TrafficLightSimulator() {
+		tlc = TrafficLightColour.RED;
+
+		thrd = new Thread(this);
+		thrd.start();
+	}
```
utopia-2, line 22: Seems like these constructors are doing most of the same work...

src/main/java/traffic/TrafficLightSimulator.java
```diff
@@ -0,0 +1,78 @@
...
+		case YELLOW:
+			tlc = TrafficLightColour.RED;
+			break;
+		case GREEN:
+			tlc = TrafficLightColour.YELLOW;
+			break;
+		case RED_YELLOW:
+			tlc = TrafficLightColour.GREEN;
+			break;
+		}
```
utopia-2, line 51: Did you look into giving `ordinal()` a shot for solving this?

xtra-answers.md
```diff
@@ -0,0 +1,15 @@
+# *Java: The Complete Answers, Twelfth Edition* <sup>(Prophet Newspaper/Wambeek Home Edition, 2015)</sup>
+
+## 10. Continued
+
+An interface with an @ symbol that isn't really an interface.
+
+## 11. What is a marker annotation?
+
+An annotation purely to mark a point in the code.
```
utopia-2, line 9: Almost - it's an annotation that contains no fields, meant purely to indicate something (such as _this method `@Override`s something from a supertype_).

src/main/java/traffic/TrafficLightColour.java
Spinodal, line 5: actually `RED_YELLOW` represents both red and yellow being on at the same time.

src/main/java/traffic/TrafficLightSimulator.java
Spinodal, line 22: copy and paste :stuck_out_tongue:

src/main/java/traffic/TrafficLightSimulator.java
Spinodal, line 51: i had a think about it but it was too late so gave up...

wait, I mean... i did it on my own as the book instructed me to...

src/main/java/traffic/TrafficLightColour.java
utopia-2, line 5: Ah, that's fair enough then - didn't realise it was the intermediary state. :+1:

PR #138, opened 2015-02-10
ncl Chapter 13

NormanLittle:


utopia-2:
:+1: for continuing Java Training :smile:.

src/com/selftest/stack/StackImpl.java
```diff
@@ -0,0 +1,59 @@
...
+
+public class StackImpl<T> implements Stack<T> {
+
+	private T[] stack;
+	private int size;
+
+	@SafeVarargs
+	public StackImpl( int size, T ... values ) {
+		stack = Arrays.copyOf( values, size );
+		size = 0;
```
utopia-2, line 17: If the user were to add values to the constructor, would they expect this to discard them?

NormanLittle, line 17: Hmm, I would imagine that would be a surprise feature.  I forgot to update the index (size)...but I am sure I _would_ have picked that up in unit testing :)

NormanLittle:
@Utopia1138 I couldn't possibly stop Java training _before_ we get to Lambdas!  Sadly, the new job is only on JDK 1.7 so I'll have to wait a wee while to use 'em in actual work.

src/com/selftest/StackDemo.java
```diff
@@ -0,0 +1,36 @@
...
+	public static void main( String[] args ) {
+		Stack<Character> stack = new StackImpl<Character>( 8 );
+		try {
+			for ( char c = 'a' ; c <= 'z' ; c++ )
+				stack.push( c );
+		}
+		catch ( StackFullException e ) {
+			System.out.printf( "%s: [%s]%n",
+			    							 e.getMessage(),
+			    							 stack.stream().map( c -> c.toString() ).collect( Collectors.joining( "," ) ) );
```
Spinodal, line 22: Wow the formatting went a bit mental here!

answers.md
```diff
@@ -0,0 +1,60 @@
...
+7)  Yes.
+```java
+static void doSomething( SomeValue<? extends Number> ) {
+}
+```
+
+8)
+```java
+public static <T> T MyGen( T t) {
+  return t;
```
NormanLittle, line 36: yay! Thanks for the 'gentleman octopus'.  Disappointed to miss the meeting this morning.

PR #140, opened 2015-02-11
spg Chapter 13

SimonGross:
Chapter 13

### DIFF
--- a/src/com/datacash/training/chapter13/QuestionTwelve.java
+++ b/src/com/datacash/training/chapter13/QuestionTwelve.java
@@ -1,0 +1,59 @@
+
+package com.datacash.training.chapter13;
+
+public class QuestionTwelve {
+
+	public static void main( String[] args ) {
+
+		try {
+
+			Stack<Character> stack = new Stack<Character>( new Character[5] );
+			// Push x2
+			stack.push( 'h' );
+			stack.push( 'e' );
+			stack.printStack();
+			// Pop
+			System.out.println( "Pop stack: " + stack.pop() );
+			stack.printStack();
+			// Push x3
+			stack.push( 'l' );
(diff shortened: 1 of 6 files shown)


src/com/datacash/training/chapter13/answers.md
```diff
@@ -0,0 +1,52 @@
...
+public class FlightSched<T, V extends T> {
+```
+
+6) `?` is the wildcard argument, which can be used when the exact type is not known/
+
+7) Yes, the wildcard argument can be bounded.
+
+8) 
+```java
+public T MyGen ( T type ) {
```
utopia-2, line 26: The question asks "_A generic method called `myGen()` has one type parameter..._", but this uses the type parameter presumably defined for the enclosing class.

alexdeas, line 26: Comments invader.

SimonGross, line 26: This is embarrassing because I implemented this correctly in some changes I recently made to HPS. The correct form is:

``` java
public <T> T MyGen ( T type ) {
```

utopia-2, line 26: Sorry @alexdeas, I didn't have anything else to review :frowning:.

src/com/datacash/training/chapter13/QuestionTwelve.java
```diff
@@ -0,0 +1,59 @@
...
+
+public class QuestionTwelve {
+
+	public static void main( String[] args ) {
+
+		try {
+
+			Stack<Character> stack = new Stack<Character>( new Character[5] );
+			// Push x2
+			stack.push( 'h' );
```
alexdeas, line 12: Autoboxing in action!

PR #141, opened 2015-02-11
Red Chapter 13 answers

utopia-2:


answers.md
```diff
@@ -0,0 +1,112 @@
...
+```java
+class FlightSched<T, E extends T> {}
+```
+
+  6) _As it relates to generics, what is the `?` and what does it do?_
+
+`?` is the wildcard, essentially allowing Java's version of an existential type.
+It allows parameterized types to be passed around without knowing the exact type
+of the parameter. If the type applies bounds to the parameter, the wildcard will
+respect that. Wildcards can also be further bounded:
```
alexdeas, line 52: Spoilers for question 7.

answers.md
```diff
@@ -0,0 +1,112 @@
...
+```
+
+  9) _Given this generic interface_
+```java
+interface IGenIF<T, V extends T> { // ...
+```
+_show the declaration of a class called `MyClass` that implements `IGenIF`._
+
+`IGenIF` - seriously? That's the name you went with? You should feel bad, and your
+editor should feel bad.
```
alexdeas, line 77: You wrote it anyway, do you feel bad?

utopia-2, line 77: I feel I sufficiently mitigated its use with this complaint.

answers.md
```diff
@@ -0,0 +1,112 @@
...
+     C. _Reliable_
+     D. _All of the above_
+
+D, in that generics _enable_ each, but writing generic code does not protect
+you from breaking each. It is possible for generic code to be unsafe, unusable,
+and unreliable.
+
+  2) _Can a primitive type be used as a type argument?_
+
+No. Java generic parameters are type erased to an `Object` or bound reference type,
```
Spinodal, line 15: But you can (kinda) if you use the wrapper class

answers.md
```diff
@@ -0,0 +1,112 @@
...
+interface IGenIF<T, V extends T> { // ...
+```
+_show the declaration of a class called `MyClass` that implements `IGenIF`._
+
+`IGenIF` - seriously? That's the name you went with? You should feel bad, and your
+editor should feel bad.
+
+Anyway:
+```java
+class MyClass implements IGenIF<Number, Long>
```
Spinodal, line 81: i think this is partially wrong red, does this not give compiler warnings?

Spinodal: i think this is partially wrong red, does this not give compiler warnings?

Spinodal, line 81: Oh wait, your cheating by using actual <parameters> rather than keeping them generic :stuck_out_tongue:

answers.md
```diff
@@ -0,0 +1,112 @@
...
+```
+
+ 10) _Given a generic class called `Counter<T>`, show how to create an object of its
+      raw type._
+
+```java
+Counter count = new Counter(0);
```
Spinodal, line 88: will passing in a `0` still autobox to `Integer` then store as object?

src/org/java/training/StackDemo.java
```diff
@@ -0,0 +1,113 @@
...
+	default Stream<T> stream() {
+		return StreamSupport.stream(spliterator(), false);
+	}
+}
+
+class StackOverflowException extends RuntimeException {
+	private static final long serialVersionUID = 1L;
+
+	public StackOverflowException(int size) {
+		super( String.format("Insertion attempt exceeds maximum stack size [%d]", size) );
```
Spinodal, line 48: you should have had the error message say: `"This post does not meet our quality standards."`

answers.md
utopia-2, line 15: That's an object, or bound reference type :smile:.

answers.md
utopia-2, line 81: It's not cheating, it fulfills the question exactly. I have declared a class that implements `IGenIF`.

answers.md
utopia-2, line 88: Oh, the integer doesn't mean anything, I just thought the `Counter` constructor could use an argument.

Spinodal, line 88: lol, but now i'm curious, i'll have to test it now

Spinodal, line 88: also, yes it does do all of the casting and this works

PR #147, opened 2015-02-18
ncl Chapter 14

NormanLittle:


alexdeas:
There was a break this week in case you're wondering where all the other pull requests are.

NormanLittle:
Ah, thanks for letting me know @alexdeas.

answers.md
```diff
@@ -0,0 +1,56 @@
...
+
+16) See [CommonFactor.java](src/com/selftest/CommonFactor.java)
+
+
+17)  
+```java
+SomeClass::new;
+```
+
+18)  java.util.function.
```
utopia-2, line 56: :+1: all good.

src/com/selftest/CommonFactor.java
```diff
@@ -0,0 +1,46 @@
...
+		return value;
+	}
+
+	boolean isFactor( int number ) {
+		return value % number == 0;
+	}
+
+	boolean hasCommonFactor( int number ) {
+		return IntStream.range( 2, value )
+		                .anyMatch( i -> isFactor( i ) && number % i == 0 );
```
utopia-2, line 26: A bit inefficient, but it does look elegant enough.

SandyThomson:
@NormanLittle - just FYI, we had a vote and are skipping the applets part of the next chapter (and just do the 'Misc Topics right at the end). As filler, we are planning to mess around with lambdas a little more.

We plan to do Chapters 16 and 17 though.

PR #153, opened 2015-03-05
spg Chapter '15'

SimonGross:
Not really answers for Week 15, but a small program showing how lambdas might be useful


SimonGross:
Closed

PR #155, opened 2015-03-12
Chapter 16 answers and code

alexdeas:
ajd chapter 16

### DIFF
--- a/src/chapter16/Checkbox.java
+++ b/src/chapter16/Checkbox.java
@@ -1,0 +1,138 @@
+package chapter16;
+
+/*
+  Based on Try This 16-1
+  
+  File comparer with a check box
+
+*/
+
+import java.awt.*;
+import java.awt.event.*;
+import javax.swing.*;
+import java.io.*;
+
+public class Checkbox implements ActionListener {
+
+  JTextField jtfFirst;  // holds the first file name
+  JTextField jtfSecond; // holds the second file name
+
+  JButton jbtnComp; // button to compare the files
+
+  JLabel jlabFirst, jlabSecond; // displays prompts
+  JLabel jlabResult; // displays results and error messages
+
+  JCheckBox checkbox;
+  JLabel checkboxResult;
+
+  Checkbox() {
+
+   // Create a new JFrame container.
+   JFrame jfrm = new JFrame("Compare Files");
+
+   // Specify FlowLayout for the layout manager.
+   jfrm.setLayout(new FlowLayout());
+
+   // Give the frame an initial size.
+   jfrm.setSize(200, 250);
+
+   // Terminate the program when the user closes the application.
+   jfrm.setDefaultCloseOperation(JFrame.EXIT_ON_CLOSE);
+
+   // Create the text fields for the file names..
+   jtfFirst = new JTextField(14);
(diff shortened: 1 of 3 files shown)


src/chapter16/MultiList.java
```diff
@@ -0,0 +1,90 @@
...
+    
+    int[] indices = jlst.getSelectedIndices();
+    
+    if ( indices.length >= 0 ) { 
+      
+      String nameList = new String();
+      for (int index : indices ) {
+        nameList += names[index] + " ";
+      }
+      nameList = nameList.trim();
```
utopia-2, line 72: If you're building up a string from an array, you're better off using a `StringBuilder` to avoid multiple temporary `String`s being allocated. This type of thing is also better suited to Java 8 streams:

``` java
Arrays.stream(indices).mapToObj(i -> names[i]).collect(Collectors.joining(" "));
```

(you can consider this equivalent to this perl:)

``` perl
join ' ', map { $names[$_] } @indices;
```

**edit**: add closing paren.

src/chapter16/MultiList.java
```diff
@@ -0,0 +1,90 @@
...
+    }
+  }  
+ 
+  public static void main(String args[]) {  
+    // Create the frame on the event dispatching thread.  
+    SwingUtilities.invokeLater(new Runnable() {  
+      public void run() {  
+        new MultiList();  
+      }  
+    });
```
utopia-2, line 87: I'm not sure why the book used this method of invoking event-dispatch thread tasks, when we've gone through lambdas and method references.
The lambda version can be simply:

``` java
SwingUtilities.invokeLater(() -> new MultiList());
```

Or even better:

``` java
SwingUtilities.invokeLater(MultiList::new);
```

src/chapter16/Checkbox.java
```diff
@@ -0,0 +1,138 @@
...
+
+   // Add the components to the content pane.
+   jfrm.add(jlabFirst);
+   jfrm.add(jtfFirst);
+   jfrm.add(jlabSecond);
+   jfrm.add(jtfSecond);
+   jfrm.add(checkbox);
+   jfrm.add(jbtnComp);
+   jfrm.add(jlabResult);
+   jfrm.add(checkboxResult);
```
utopia-2, line 74: I wish `JFrame` had a variadic version of `add` to reduce this type of boilerplate. In modern Java it'd be a two line method (or one line if you hate operator chaining):

``` java
public JFrame add(Component... components) {
  Arrays.stream(components).forEachOrdered(this::add);
  return this;
}
```

src/chapter16/MultiList.java
alexdeas, line 72: Unfortunately since I read this book I know next to nothing about streams. That's pretty much what I wanted to do though.

Also syntax error.

src/chapter16/MultiList.java
alexdeas, line 87: I presumed they wanted to show off anonymous classes without really telling you much about them.

src/chapter16/MultiList.java
utopia-2, line 72: You can't expect my comments to compile :wink:.

PR #156, opened 2015-03-18
ajd chapter 17

alexdeas:
ajd chapter 17

### DIFF
--- a/answers.md
+++ b/answers.md
@@ -1,0 +1,31 @@
+1) ``javafx``
+
+2) ``Stage`` and ``Scene``? Seems too obvious.
+
+3) A scene graph is composed of ''nodes''.
+
+4) The base class for all nodes is ``Node``.
+
+5) ``Application``
+
+6) The lifecycle methdos are ``init()``, ``start()`` and ``stop()``
+
+7) The stage must be instantiated in the ``start()`` method.
+
+8) True.
+
+9) Labels and buttons are supported by the ``Label`` and ``Button`` classes, without any silly prefixes.
+
+10) See JavaFXEventDemo.java
+
+11) ``CheckBox`` (really?)
+
+12) False!
+
+13) See JavaFXFC.java
+
+14) See EffectsAndTransformsDemo.java
+
+15) Ran out of time
+
+16) TBC
--- a/src/chapter17/EffectsAndTransformsDemo.java
+++ b/src/chapter17/EffectsAndTransformsDemo.java
@@ -1,0 +1,134 @@
+package chapter17;
+
+import javafx.application.*;  
+import javafx.scene.*;  
+import javafx.stage.*;  
+import javafx.scene.layout.*;  
+import javafx.scene.control.*;  
+import javafx.event.*;  
+import javafx.geometry.*;  
+import javafx.scene.transform.*; 
+import javafx.scene.effect.*; 
+  
+public class EffectsAndTransformsDemo extends Application {  
+  
+  double angle = 0.0; 
+  double scaleFactor = 0.4; 
+  double blurVal = 1.0; 
+ 
+  // Create initial effects and transforms. 
+  Reflection reflection = new Reflection(); 
+  BoxBlur blur = new BoxBlur(1.0, 1.0, 1); 
+  Rotate rotate = new Rotate(); 
+  Scale scale = new Scale(scaleFactor, scaleFactor); 
+ 
+  // Create push buttons.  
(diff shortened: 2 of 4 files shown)


answers.md
```diff
@@ -0,0 +1,31 @@
+1) ``javafx``
+
+2) ``Stage`` and ``Scene``? Seems too obvious.
```
utopia-2, line 3: Sounds like a _good_ thing for an API :boar:.

src/chapter17/EffectsAndTransformsDemo.java
```diff
@@ -0,0 +1,134 @@
...
+  
+  // Blur the rotate button too
+  BoxBlur rotateBlur = new BoxBlur(5.0, 5.0, 2); 
+  boolean doBlur = true;
+ 
+  public static void main(String[] args) {  
+  
+    // Start the JavaFX application by calling launch().  
+    launch(args);    
+  }
```
utopia-2, line 40: The latest eclipse seems to have run configurations for JavaFX applications (so you wouldn't need `main`).

src/chapter17/JavaFXEventDemo.java
```diff
@@ -0,0 +1,62 @@
...
+
+import javafx.application.*; 
+import javafx.scene.*; 
+import javafx.scene.layout.*; 
+import javafx.scene.control.*; 
+import javafx.geometry.*; 
+ 
+public class JavaFXEventDemo extends Application { 
+ 
+  Label response;
```
utopia-2, line 15: This needn't be an instance variable. :crocodile:

src/chapter17/JavaFXEventDemo.java
```diff
@@ -0,0 +1,62 @@
...
+ 
+    // Create two push buttons. 
+    Button btnUp = new Button("Run"); 
+    Button btnDown = new Button("Exit"); 
+ 
+    // Handle the action events for the Run button. 
+    btnUp.setOnAction( (ae) -> response.setText( "You pressed Run" ) ); 
+ 
+    // Handle the action events for the Exit button. 
+    btnDown.setOnAction( (ae) -> Platform.exit() );
```
utopia-2, line 53: For single arg lambdas, you can leave out the parentheses, if you like (or if you're in some kind of parenthesis famine): :cold_sweat: 

``` java
ae -> Platform.exit()
```

src/chapter17/JavaFXFC.java
```diff
@@ -0,0 +1,104 @@
...
+         if(i != j)
+           result.setText("Files are not the same.");
+         else
+           result.setText("Files are the same.");
+      } catch(IOException exc) {
+        result.setText("File Error");
+      }
+        
+      
+    } );
```
utopia-2, line 90: This seems like a lot of body for one lambda - might be worth pulling it out to a method. You lose the advantage of lexical closure capture, but you could make it non-gui specific:

``` java
public String compareFiles(String patha, String pathb);
```

``` java
compare.setOnAction( e -> result.setText(compareFiles(file1.getText(), file2.getText())) );
```

answers.md
```diff
@@ -0,0 +1,31 @@
...
+
+12) False!
+
+13) See JavaFXFC.java
+
+14) See EffectsAndTransformsDemo.java
+
+15) Ran out of time
+
+16) TBC
```
utopia-2, line 31: :shipit:

src/chapter17/EffectsAndTransformsDemo.java
alexdeas, line 40: That's what I get from copying the example code.

src/chapter17/JavaFXEventDemo.java
alexdeas, line 53: What if I'm into superfluous parentheses?

src/chapter17/JavaFXFC.java
alexdeas, line 90: I had a method originally, but took it out as I was only using it once. Hate having single use methods.

utopia-2, line 90: Whaaat, a well named single use method that does one conceptual task is far better than a wall of code that interleaves behavior. That said, a single-use _public_ method would be awful.

src/chapter17/JavaFXEventDemo.java
utopia-2, line 53: Sure, if you're into it, who am I to judge? :kissing_heart:

